PR #8, opened 2016-12-01
Add Raw IMU and RC in messages

bugobliterator:


pavel-kirienko:
Thanks for the submission.

- Why do you have both integrated and averaged quantities in the IMU message? Average can be deduced from integrated if needed.

- You stated that the quantities of angular velocity and acceleration are in SI units, which is correct. Yet the fields are suffixed with `_raw`, why?

- What is the exact purpose of the RC message? We already have actuator and ESC setpoint messages.

bugobliterator:
- Why do you have both integrated and averaged quantities in the IMU message? Average can be deduced from integrated if needed.
*_raw message does not represent averaged quatity but latest point sample. The usage could be for accurate sensor calibration and if someone's AHRS requires point samples instead of intergrals.

- You stated that the quantities of angular velocity and acceleration are in SI units, which is correct. Yet the fields are suffixed with _raw, why?
I felt that since the samples are point i.e. unprocessed except unit transformation it is justified to use that term.

- What is the exact purpose of the RC message? We already have actuator and ESC setpoint messages.
Its an RC in message i.e. message received from radio transmitter. I found a requirement of such a message when I am working with a redundant autopilot system, where all the autopilots are supplied with info like RC and IMU on CAN bus.

pavel-kirienko:
> *_raw message does not represent averaged quatity but latest point sample. The usage could be for accurate sensor calibration and if someone's AHRS requires point samples instead of intergrals.

OK.

>> You stated that the quantities of angular velocity and acceleration are in SI units, which is correct. Yet the fields are suffixed with _raw, why?

> I felt that since the samples are point i.e. unprocessed except unit transformation it is justified to use that term.

I understand your intention, but I don't think "raw" is the correct word to use here. Let's call it either:

* point
* last
* no suffix at all
* I'm open to further suggestions

It probably would be beneficial to have fields for error covariance estimates for gyro and accel samples, as this information can be efficiently used by state estimators. Do you agree?

>> What is the exact purpose of the RC message? We already have actuator and ESC setpoint messages.

> Its an RC in message i.e. message received from radio transmitter. I found a requirement of such a message when I am working with a redundant autopilot system, where all the autopilots are supplied with info like RC and IMU on CAN bus.

Having a message carrying a set of arbitrary control inputs could be useful, but there is no reason to constrain it to RC only. For instance, it could be used to provide control input from a companion computer. What I propose is to rename the message from `Radioin` (should have been `RadioIn` anyway) into something more generic, i.e. `ControlInput` or something of that effect. The namespace should be renamed accordingly.

I am open to other suggestions.

bugobliterator:
@pavel-kirienko I've made the said changes and some additions. Please review.

uavcan/equipment/ahrs/1002.RawIMU.uavcan
```diff
@@ -0,0 +1,21 @@
...
+#
+
+#
+# rad/sec
+#
+uavcan.Timestamp gyro_timestamp
+float16[3] gyro_point
+float16[3] gyro_integral
+float16 gyro_integral_dt
+float16[<=9] gyro_covariance
```
pavel-kirienko: Missing alignment here - the field `gyro_covariance` should be prepended with `void4` in order to maintain byte alignment for all elements of the array.

uavcan/equipment/ahrs/1002.RawIMU.uavcan
```diff
@@ -0,0 +1,21 @@
...
+
+#
+# m/sec^2
+#
+uavcan.Timestamp accel_timestamp
+float16[3] accel_point
+float16[3] accel_integral
+float16 accel_integral_dt
+float16[<=9] accel_covariance
```
pavel-kirienko: Missing trailing newline. Note that no alignment is needed here, because tail array optimization will elide the size field.

uavcan/equipment/control/1110.ControlInput.uavcan
```diff
@@ -0,0 +1,31 @@
+#
+# This message represents array of arbitrary control input
+#
+
+#
+# Nested type.
+# Single actuator command.
+#
+uint8 controller_id
```
pavel-kirienko: What would be the use case for this field?

uavcan/equipment/control/1110.ControlInput.uavcan
```diff
@@ -0,0 +1,31 @@
...
+
+#
+# Controlled Device Type
+#
+uint8 DEV_TYPE_UNDEF            = 0
+uint8 DEV_TYPE_AUTOPILOT        = 1
```
pavel-kirienko: Same question about these constants - why do we need them, and where should these values be stored? I don't see any corresponding field like `dev_type` nearby.

uavcan/equipment/control/1110.ControlInput.uavcan
bugobliterator: For systems with redundant controllers, like the one I am currently working on.

uavcan/equipment/ahrs/1002.RawIMU.uavcan
```diff
@@ -0,0 +1,23 @@
+#
+# Raw IMU data with timestamps
+#
+
+
```
pavel-kirienko: Extra newlines

uavcan/equipment/ahrs/1002.RawIMU.uavcan
```diff
@@ -0,0 +1,23 @@
...
+#
+# rad/sec
+#
+uavcan.Timestamp gyro_timestamp
+float16[3] gyro_raw
+float16[3] gyro_integral
+void4
+float16 gyro_integral_dt
```
pavel-kirienko: I noticed that you removed the field `float16[<=9] gyro_covariance`. If you chose to remove it, you should also have removed the alignment field `void4`. I suggest to keep the covariance field though, probably in the form of a single 6x6 matrix at the end of the message. In that case it will be possible to convey covariance information between acceleration and angular velocity measurements. The right syntax is `float16[<=36] gyro_accel_covariance`. See http://uavcan.org/Specification/5._Application_level_conventions/#matrix-representation.

pavel-kirienko:
@bugobliterator nudge!

bugobliterator:
@pavel-kirienko Sorry, I lost track of this PR. I have added the changes you requested. Feel free to pull it in.

pavel-kirienko:
Superseded by https://github.com/UAVCAN/dsdl/pull/27

Issue #7, opened 2016-09-28
GNSS: Limitations of the current uavcan.equipment.gnss.Fix message


WickedShell:
Status doesn't allow for reporting any fix types better then 3D fix. This is a problem as there are significant differences between 3D, DGPS, RTK Float, RTK Fixed types. (As well as not allowing for future reporting of a fixed base status).

The accuracy estimates can be used for comparing which GPS to use, however there are several issues with trying to use it in this manner:
- Different GPS units will have different fundamental accuracies in these modes, which makes it impossible to use the accuracy to discriminate which mode the GPS is in
- Accuracies from GPS units aren't reported within the MAVLink protocol, which makes it impossible to determine with certainty which mode the GPS is in
- There is no support for reporting a GPS is being used as a fixed base to generate RTCM data (IE a u-blox M8P)

pavel-kirienko:
Fix type was not included in the message because it was intended that the final use of the fix type is to detect the quality, hence precision, of the fix, and this function was delegated to the error covariance estimate, provided in the matrix field [`position_covariance`](http://uavcan.org/Specification/7._List_of_standard_data_types/#fix).

I now realize that this approach, indeed, did not account for all viable use cases. A possible solution is to utilize one of the reserved fields in the message (`void5` or `void4`), converting them into an integer field defining the fix type, and freezing the message DSDL signature using the approach outlined here https://github.com/UAVCAN/dsdl/issues/4. Since we're going to re-use a reserved field, the binary compatibility should not be harmed - receivers that use old definitions will just continue to ignore the new field, and transmitters that use the old format will keep reporting zero in the new field, which should be declared to be a functinally-neutral value, e.g. indicating that the fix type is undefined or something of that effect.

What exact fix type codes would you propose to define? Keep in mind that there is already the `status` field which we can't touch, therefore the new type definitions should be orthogonal to it.

WickedShell:
Since we already have the status field, the additional fix types I'd recommend would be:
DGPS/SBAS
RTK Float
RTK Fixed
Base (or static position)

There is a bit of a debate if DGPS should be split into two parts (DGPS and SBAS). Technically the two are different, but given that few people leverage a DGPS system other then SBAS, I'm not sure it makes sense. (For some hardware like the u-blox chips, it takes more work to distinguish between the two types of fixes). Given that the current MAVLink protocol treats them as identical, and overall DGPS usage is rarer then the other modes I think reporting it as a combination is probably fine.

[...]

PR #10, opened 2017-01-29
Addition of ECEF coordinate frame for GNSS-message

### LINKED ISSUE #9: GNSS: 1060.Fix.uavcan ECEF instead of LLA
Some GNSS receivers output ECEF coordinates instead of LLA. Are there plans to provide support for these fields in a GNSS fix message?

glennib:
Added draft for a new GNSS-message containing ECEF coordinate representation: 1063.Fix2.uavcan. In addition, a new field for a 3x3 covariance matrix is included: pos_vel_covariance. This describes the covariance between position and velocity. Since covariance matrices are symmetric, only the upper-right quadrant of the 6-by-6 matrix is required.

Fixes #9

### DIFF
--- a/uavcan/equipment/gnss/1063.Fix2.uavcan
+++ b/uavcan/equipment/gnss/1063.Fix2.uavcan
@@ -1,0 +1,64 @@
+#
+# GNSS ECEF navigation solution with uncertainty.
+#
+
+uavcan.Timestamp timestamp         # Global network-synchronized time, if available, otherwise zero
+
+#
+# Time solution.
+# Time standard (GPS, UTC, TAI, etc) is defined in the field below.
+#
+uavcan.Timestamp gnss_timestamp    # GNSS timestamp, if available, otherwise zero
+
+#
+# Time standard used in the GNSS timestamp field.
+#
+uint3 GNSS_TIME_STANDARD_NONE = 0  # Time is unknown
+uint3 GNSS_TIME_STANDARD_TAI  = 1
+uint3 GNSS_TIME_STANDARD_UTC  = 2
+uint3 GNSS_TIME_STANDARD_GPS  = 3
+uint3 gnss_time_standard
+
+void5   # Reserved space
+
+#
+# If known, the number of leap seconds allows to perform conversions between some time standards.
+#
+uint8 NUM_LEAP_SECONDS_UNKNOWN = 0
+uint8 num_leap_seconds
+
+#
+# Position and velocity solution
+#
+int37 longitude_deg_1e8            # Longitude degrees multiplied by 1e8 (approx. 1 mm per LSB)
+int37 latitude_deg_1e8             # Latitude degrees multiplied by 1e8 (approx. 1 mm per LSB on equator)
+int27 height_ellipsoid_mm          # Height above ellipsoid in millimeters
+int27 height_msl_mm                # Height above mean sea level in millimeters
+
+float16[3] ned_velocity            # NED frame (north-east-down) in meters per second
+
+#
+# Fix status
+#
+uint6 sats_used
+
+uint2 STATUS_NO_FIX    = 0
+uint2 STATUS_TIME_ONLY = 1
+uint2 STATUS_2D_FIX    = 2
+uint2 STATUS_3D_FIX    = 3
+uint2 status
+
+#
+# Precision
+#
+float16 pdop
+
+float16[<=9] position_covariance   # m^2
+float16[<=9] velocity_covariance   # (m/s)^2
+float16[<=9] pos_vel_covariance    # m^2 / s - Covariance 
+
+#
+# Position solution in ECEF
+#
+
+ECEFPosition[<=1] ecef_position
--- a/uavcan/equipment/gnss/1063.Fix2.uavcan
+++ b/uavcan/equipment/gnss/1063.Fix2.uavcan
@@ -19,7 +19,7 @@ uint3 GNSS_TIME_STANDARD_UTC  = 2
 uint3 GNSS_TIME_STANDARD_GPS  = 3
 uint3 gnss_time_standard
 
-void5   # Reserved space
+void13   # Reserved space
 
 #
 # If known, the number of leap seconds allows to perform conversions between some time standards.
@@ -49,7 +49,7 @@ uint2 STATUS_3D_FIX    = 3
 uint2 status
 
 #
-# Mode status
+# GNSS Mode
 # 
 uint3 MODE_SINGLE      = 0
 uint3 MODE_DGPS        = 1
@@ -57,30 +57,27 @@ uint3 MODE_RTK         = 2
 uint3 MODE_PPP         = 3
 uint3 mode
 
-void1   # Reserved space
-
 #
-# Submode status
+# GNSS Sub mode
 #
 
-uint3 SUB_MODE_DGPS_OTHER    = 0
-uint3 SUB_MODE_DGPS_SBAS     = 1
+uint5 SUB_MODE_DGPS_OTHER    = 0
+uint5 SUB_MODE_DGPS_SBAS     = 1
 
-uint3 SUB_MODE_RTK_FLOAT     = 0
-uint3 SUB_MODE_RTK_FIXED     = 1
+uint5 SUB_MODE_RTK_FLOAT     = 0
+uint5 SUB_MODE_RTK_FIXED     = 1
 
-uint3 sub_mode
-
-void1   # Reserved space
+uint5 sub_mode
 
 #
 # Precision
 #
 float16 pdop
 
-float16[<=6] position_covariance   # m^2
-float16[<=6] velocity_covariance   # (m/s)^2
-float16[<=9] pos_vel_covariance    # m^2 / s - Covariance of position and velocity.
+float16[<=21] covariance    # Position and velocity covariance. The upper-right
+                            # triangle of a 6x6 symmetric matrix, represented in
+                            # row-major order. Units are m^2 for position, (m/s)^2
+                            # for velocity and m^2/s for position/velocity.
 
 #
 # Position solution in ECEF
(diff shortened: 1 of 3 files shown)


uavcan/equipment/gnss/1063.Fix2.uavcan
```diff
@@ -0,0 +1,64 @@
...
+#
+# Fix status
+#
+uint6 sats_used
+
+uint2 STATUS_NO_FIX    = 0
+uint2 STATUS_TIME_ONLY = 1
+uint2 STATUS_2D_FIX    = 2
+uint2 STATUS_3D_FIX    = 3
+uint2 status
```
WickedShell: Unless you are always planning to send the FIX message as well you need to incorporate some form of reporting DGPS/RTK etc. which is what #7 was about.

glennib:
Please review the latest changes. @WickedShell @pavel-kirienko

uavcan/equipment/gnss/1063.Fix2.uavcan
```diff
@@ -0,0 +1,92 @@
+#
+# GNSS ECEF navigation solution with uncertainty.
```
pavel-kirienko: GNSS ECEF **and LLA** navigation solution with uncertainty.

Please mark the old Fix message deprecated (in the header comments).

uavcan/equipment/gnss/1063.Fix2.uavcan
```diff
@@ -0,0 +1,92 @@
...
+void1   # Reserved space
+
+#
+# Precision
+#
+float16 pdop
+
+float16[<=9] position_covariance   # m^2
+float16[<=9] velocity_covariance   # (m/s)^2
+float16[<=9] pos_vel_covariance    # m^2 / s - Covariance of position and velocity.
```
pavel-kirienko: Clever.

Now, do we want to ensure byte alignment  for each array? If we do, we'll need to add `void4` before each array. If we don't, we should keep everything intact.

My proposition is to ensure byte alignment, since that should enable faster serialization/deserialization, and provide us with 12 spare bits for being extra future proof.

uavcan/equipment/gnss/ECEFPosition.uavcan
```diff
@@ -0,0 +1,17 @@
...
+# coordinate system which rotates with the earth. The origin (0,0,0) is
+# located at the center of the earth. The x-axis is a vector pointing from
+# the origin with positive direction towards 0 degrees latitude and
+# longitude (equator, at the prime meridian). The z-axis is a vector
+# pointing from the origin towards the north-pole. The y-axis completes a
+# right-handed coordinate system.
+
+int34 x_mm                         # X-axis coordinate in mm
+int34 y_mm                         # Y-axis coordinate in mm
+int34 z_mm                         # Y-axis coordinate in mm
```
pavel-kirienko: We could sacrifice one more byte to make the message definition suitable for use in space. This is what I propose:

```
int36[3] xyz_mm
void4
```

uavcan/equipment/gnss/1063.Fix2.uavcan
pavel-kirienko: Seems to be fixed now.

uavcan/equipment/gnss/1063.Fix2.uavcan
```diff
@@ -0,0 +1,92 @@
...
+#
+uint3 SUB_MODE_SINGLE        = 0
+
+uint3 SUB_MODE_DGPS_OTHER    = 0
+uint3 SUB_MODE_DGPS_SBAS     = 1
+
+uint3 SUB_MODE_RTK_FLOAT     = 0
+uint3 SUB_MODE_RTK_FIXED     = 1
+
+uint3 SUB_MODE_PPP           = 0
```
pavel-kirienko: This is not needed - an enumeration of one element seems to be useless. Please find a similar issue above with `SUB_MODE_SINGLE`.

uavcan/equipment/gnss/ECEFPosition.uavcan
mhkabir: +1

uavcan/equipment/gnss/1063.Fix2.uavcan
glennib: I'm not sure I understand what you mean by byte alignment here. Please elaborate

I also have another suggestion. Since covariance matrices are symmetric, I suggest changing `float16[<=9] position_covarnace` and `float16[<=9] velocity_covarnace` to `float16[<=6] position_covarnace` and `float16[<=9] velocity_covarnace`, respectively. This will include all required information, and save 12 bytes.

uavcan/equipment/gnss/1063.Fix2.uavcan
glennib: When the `mode` field is in PPP or single, `sub_mode` is not used. Shouldn't there be an indicator somewhere?

pavel-kirienko:
This looks alright.

Per https://github.com/UAVCAN/dsdl/issues/7, the field `sub_mode` should be 4 bits large. Since we're redefining the whole message, I would propose to use this opportunity to make it 5 bits wide using the two reserved bits before and after it, and increase the first `void5` to `void13`, to have some space reserved for future bit fields, shall we need any.

@WickedShell do you have any further propositions?

pavel-kirienko:
This has already been done.

mhkabir:
There are 3 separate matrices representing the position variance, velocity variance and the vel-pos cross covariance. Why not just have one?

pavel-kirienko:
It saves 18 bytes of the worst case message size. Although I'm OK with a
single 6x6 matrix also - both approaches are identical bus traffic wise.

glennib:
I support the change ot a 6x6 covariance matrix, but I would represent it as a 21 element float16 array. Since the covariance matrix is symmetric, we save 30 bytes.

uavcan/equipment/gnss/1063.Fix2.uavcan
glennib: @pavel-kirienko Could you elaborate on this byte alignment thing?

mhkabir:
@glennib Awesome!
I was just about to suggest that we transport the upper triangular. :)

pavel-kirienko:
Folks, no need to complicate things. 6x6 and its triangle are identical bandwidth-wise. Please see http://uavcan.org/Specification/5._Application_level_conventions/#matrix-representation

uavcan/equipment/gnss/1063.Fix2.uavcan
```diff
@@ -0,0 +1,86 @@
...
+
+#
+# Precision
+#
+float16 pdop
+
+float16[<=21] covariance    # Position and velocity covariance. The upper-right
+                            # triangle of a 6x6 symmetric matrix, represented in
+                            # row-major order. Units are m^2 for position, (m/s)^2
+                            # for velocity and m^2/s for position/velocity.
```
pavel-kirienko: In the course of pursuing clarity over memory footprint, I suggest to replace it with `float16[<=36] covariance`, or keep the previous approach with three matrices. Thanks to the [matrix packing rules explained in the specification](http://uavcan.org/Specification/5._Application_level_conventions/#matrix-representation), no traffic overhead will be introduced.

glennib: I'm not completely sure of what you mean. Won't an array with 36 float elements use 36*2 bytes, and an array with 21 float elements use 21*2 bytes bandwidth? Are you suggesting the parser understands that it's a symmetric matrix and removes the 15 redundant elements? How does it then differ between a 21 element array, 6x6 full matrix, and a 6x6 upper triangular matrix?

There might be something I'm missing here.

pavel-kirienko: > Won't an array with 36 float elements use 36 * 2 bytes, and an array with 21 float elements use 21 * 2 bytes bandwidth?

Yes. However, the protocol specifies that special matrix forms (namely scalar, diagonal, and symmetric) can be compressed in order to save bandwidth. So, when a node wants to transmit a diagonal matrix, it can do so in two ways:

* transmit the full matrix, using N<sup>2</sup> elements;
* transmit only the upper-right triangle, using only ((1+N)*N)/2 elements.

The receiving nodes will know N from the data type definition, and therefore they will be able to deduce how to handle the received matrix. Namely:

* if an array of N<sup>2</sup> elements is received, no unpacking is necessary;
* if an array of ((1+N)*N)/2 elements is received, it is mirrored about the diagonal;
* an array of N elements is assigned to the diagonal of a zero NxN matrix;
* an array of 1 element is handled like [1, 1, 1] for the case above;
* an empty array is transformed into an empty matrix.

Libuavcan does provide several convenience methods that automate packing/unpacking: https://github.com/UAVCAN/libuavcan/blob/master/libuavcan/include/uavcan/marshal/array.hpp#L954-L1084. In other libraries the developer will have to pack/unpack manually.

pavel-kirienko: Typo in the last clause: "empty matrix" is supposed to mean "zero matrix".

glennib: OK gotcha. Thanks for the clarification. Will do an update.

glennib:
Changed the number of array elements to 36. Please review.

uavcan/equipment/gnss/1063.Fix2.uavcan
```diff
@@ -0,0 +1,85 @@
...
+uint5 SUB_MODE_RTK_FIXED     = 1
+
+uint5 sub_mode
+
+#
+# Precision
+#
+float16 pdop
+
+float16[<=36] covariance    # Position and velocity covariance. Units are
```
pavel-kirienko: We need `void2` before this field to ensure byte alignment. Alternatively, we can extend `mode` and `sub_mode` by 1 bit each and move `pdop` below `covariance`.

pavel-kirienko:
I applied the latest changes myself and merged; authoring information has been preserved. Thanks @glennib !

PR #12, opened 2017-01-30
Added message 1012.AdvancedStatus.uavcan.

### LINKED ISSUE #11: Actuator: Advanced actuator status message
My company uses Dynamixel Pro servos for actuation of thrust vectoring in a UAV. These servos have advanced status fields like _temperature_, _voltage_, and some other fields. I want to create a new actuator status message: `1012.AdvancedStatus.uavcan`, which contains these fields.

glennib:
Advanced actuator status message containing temperature and voltage fields, as well as an actuator-specific error code field. Fixes #11


pavel-kirienko:
Thanks Glenn.

First of all, I would propose to *replace* the current actuator status message with your new definition, rather than extending it with an additional message. Let us call it `Status2`, following the example set with `uavcan.equipment.gnss.Fix2`. Please also mark the old Status message deprecated (in the comments).

Concerning the new fields, could you please clarify the purpose of the `voltage` field? Is it supposed to contain the supply voltage of the servo? Do you think we could remove it from the actuator status message in favor of using the more generic [`uavcan.equipment.power.CircuitStatus`](http://uavcan.org/Specification/7._List_of_standard_data_types/#circuitstatus)? Note that ESC does include the voltage field, that is because in the context of an electric drive, voltage has a vital meaning, so it was decided to include it into the main status message instead of using the generic message. This is not the case with actuators, because the same interface could be applied to a non-electrically driven actuator, e.g. a hydraulically or pneumatically driven one.

glennib:
The circuit status message seems suitable for this purpose. I hadn't done my research. I will upload a new suggestion.

glennib:
One of the reasons I wanted to put these fields in a separate message, is that the update rate of e.g. temperature doesn't need to be as high as the other fields. In a dynamic system, one might want to know the position of an actuator at 200 Hz, while temperature only at 1 Hz.

pavel-kirienko:
> In a dynamic system, one might want to know the position of an actuator at 200 Hz, while temperature only at 1 Hz.

This is a fair point. On one hand, the extra 2 bytes of data will not create much of a noticeable overhead, especially for the upcoming CAN-FD networks (which we undoubtedly will support in a backward compatible manner). On the other hand, this use case might not be unique for actuators. I'm wondering if we should factor out another generic message, like the above mentioned `CircuitStatus`, for purposes of temperature (and possibly something else) monitoring, e.g. `DeviceTemperature`?

glennib:
Separating out the temperature in its own message might be suitable for multiple purposes, so this might be a good idea. The `Device Temperature` message should contain some identification field as well. The identification field in the actuator messages is named `actuator_id`, therefore specific to actuators. Should we perhaps change those out with a more generic `device_id`, where the id-range would be shared across all _devices_?

pavel-kirienko:
It would appear that we both agree that in the case of actuators, `actuator_id`, `DeviceTemperature.device_id`, and `CircuitStatus.circuit_id` are all supposed to match. The question is, do we require the identification fields to be named identically as well? I expect that to hinder flexibility somewhat. One could imagine a use case where a component whose temperature should be measured cannot be sensibly considered a dedicated `device`, same stands for CircuitStatus. Besides, changing everything to `device_id` would require us to introduce breaking changes into other type definitions. Therefore I suggest to keep the current naming.

glennib:
Please review the latest change. I removed the status2 message and introduced a message named Temperature in the Device folder. How's this?

uavcan/equipment/actuator/1011.Status.uavcan
```diff
@@ -1,6 +1,7 @@
 #
 # Generic actuator feedback, if available.
 # Unknown fields should be set to NAN.
+# This message is deprecated. Use the newer 1012.Status2.uavcan message.
```
pavel-kirienko: This is no longer true.

uavcan/equipment/device/1100.Temperature.uavcan
```diff
@@ -0,0 +1,11 @@
+#
+# Generic device temperature
+#
+
+uint16_t device_id
+
+float16 temperature
+
+uint8 ERROR_FLAG_OVERTEMPERATURE = 1
+uint8 ERROR_FLAG UNDERTEMPERATURE = 2
```
pavel-kirienko: My English is rusty so I could be wrong, but I suspect that better terms would be "overheating" and ["overcooling"](https://en.wiktionary.org/wiki/overcooling).

uavcan/equipment/device/1100.Temperature.uavcan
```diff
@@ -0,0 +1,11 @@
+#
+# Generic device temperature
+#
+
+uint16_t device_id
+
+float16 temperature
```
pavel-kirienko: While we're at it, it would help to add a comment indicating that the temperature should be in Kelvin.

pavel-kirienko:
Note that data type ID 1100 is already used, please pick a different one.

glennib:
Changed added. Please review.

pavel-kirienko:
:+1:

PR #13, opened 2017-03-05
equipment.actuator: added PWM uS pulse length type of command

EShamaev:
Hobby grade servos are taking the PWM signal as a position


pavel-kirienko:
Thank you Eugene for this contribution.

PWM us is not a suitable unit of measurement for controlling any mechanical equipment. The publishing node is supposed to convert the setpoint into one of the SI dimensional units supported by this message definition, and the receiving node is supposed to interpret it in whatever way can be understood by the connected servo. In the specific case of APM one perhaps should convert PWM us into `COMMAND_TYPE_UNITLESS`.

We don't want to keep alive inherently broken interfaces, do we?

EShamaev:
I agree that PWM is not good for control, but 99% of hobby and also lots of professional UAVs use it and it already became standard no matter if want we this or not.
Anyway, COMMAND_TYPE_UNITLESS is the solution I used now for APM to prevent double conversion at APM itself and servo receiver.

PR #15, opened 2017-03-11
Added new message for vendor specific status

EShamaev:
Added new message for PWM servo controlling hardware.
It has 2 power supplies, internal voltage regulator for servo and backup PWM input.
It reports input voltages, output voltage, temperature and error flags once per second.

### DIFF
--- a/uavcan/equipment/actuator/1020.SmartServoStatus.uavcan
+++ b/uavcan/equipment/actuator/1020.SmartServoStatus.uavcan
@@ -1,0 +1,20 @@
+#
+# SmartServo PWM servo converter feedback
+# Aerotech LLC (c) 2017
+#
+
+uint8 actuator_id
+
+# Voltage and temperature
+float16[2] in_voltage        # 2 input voltages
+float16 out_voltage          # 1 output voltage
+float16 temperature          # Kelvins
+
+# status bits
+uint8 ERROR_FLAG_VOLTAGE_IN_1_CRITICAL = 1
+uint8 ERROR_FLAG_VOLTAGE_IN_2_CRITICAL = 2
+uint8 ERROR_FLAG_VOLTAGE_OUT_CRITICAL = 4
+uint8 ERROR_FLAG_PWM_IN_MISSING = 8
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
Thank you Eugene.

I am concerned by the lack of generality in your proposed data type definition. The purpose of the multiple voltage and temperature fields can be served well by means of standard messages `uavcan.equipment.power.CircuitStatus` (you can set `current` to NaN if it is irrelevant) and `uavcan.equipment.device.Temperature`. The general status information can be equally well reported via the `vendor_specific_status_code` field of the NodeStatus message.

That said, I am beginning to see what caused such a misunderstanding. This PR and a couple of similar earlier discussions lead me to suspect that there might be a use case for fixed data type ID allocations for truly vendor-specific messages that lack generality. Similar to the ways of the USB consortium, we might explore the following idea: dedicating a fixed range of data type ID for purposes of non-general, truly vendor-specific messages, where said messages would be allocated within the vendor's specific namespaces, e.g. `aerotech.smart_servo.Status` in your specific case. The UAVCAN team would charge a small fee per every allocated data type ID that would guarantee that no dead and unused messages retain in the set.

EShamaev:
The reason I made this type is to save number of packets over the bus. Of course, 3 CircuitStatus and temperature can be send, but it's several times the overhead.
I fully agree to you next point, however in regards to charging fee, I think that UAVCAN has not had such widespread use. However all will depend on the amount of this fee.

pavel-kirienko:
Never mind the number of packets. I trust you don't need to update temperature and voltage 100+ times per second, so the generic solution should work fine. Please close this ticket if the generic approach would work for you; if not, please elaborate on your concerns and we'll try to work out a better solution for you.

Re fee, its only purpose is to prevent appearance of unused definitions, rather than attempts at monetizing. We could offer riddles to solve instead, but that would be way too unorthodox perhaps.

EShamaev:
Alright, I will use standard types for now.
Let me know when you come up with value of vendor type, I might be interested.

PR #17, opened 2017-03-28
Common aircraft lights IDs

EShamaev:
General common types of lights to be encountered in private, commercial and military planes, helicopters and UAVs.


pavel-kirienko:
:+1:

PR #19, opened 2017-04-11
Position+Attitude complex message

EShamaev:
I have added baro height and linear_acceleration to the proposed message

### DIFF
--- a/uavcan/navigation/2000.GlobalNavigationSolution.uavcan
+++ b/uavcan/navigation/2000.GlobalNavigationSolution.uavcan
@@ -1,0 +1,41 @@
+#
+# Inertial data and orientation in body frame with fused location
+#
+
+uavcan.Timestamp timestamp
+
+float64 longitude                   # angular degree
+float64 latitude                    # angular degree
+float32 height_ellipsoid            # meter
+float32 height_msl                  # meter
+float32 height_baro                 # meter
+float16 qnh                         # hPa
+float32 height_agl                  # meter (with radar altimeter or LIDAR)
+
+float32[4] orientation_xyzw         # rotation quaternion, zero yaw at north
+
+#
+# Column order:
+#   longitude       [meter^2]
+#   latitude        [meter^2]
+#   height (worse)  [meter^2]
+#   roll angle      [radian^2]
+#   pitch angle     [radian^2]
+#   yaw angle       [radian^2]
+#
+float16 pose_covariance[<=36]
+
+float32[3] linear_velocity_ned      # meter/second
+float32[3] angular_velocity         # rad/sec
+float16[3] linear_acceleration      # m/s^2
+
+#
+# Column order:
+#   north velocity  [(meter/second)^2]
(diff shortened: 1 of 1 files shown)


EShamaev:
Added distance to ground if aircraft is equipped with LIDAR or radalt

uavcan/equipment/ahrs/1002.PositionAttitude.uavcan
```diff
@@ -0,0 +1,40 @@
...
+#
+
+uavcan.Timestamp timestamp
+
+float64 longitude                   # angular degree
+float64 latitude                    # angular degree
+float32 height_ellipsoid            # meter
+float32 height_msl                  # meter
+float32 height_baro                 # meter
+float32 ground_distance             # meter (with radar altimeter or LIDAR)
```
pavel-kirienko: Shall we rename `ground_distance` into `height_agl` for consistency?

Also, that's a lot of heights, but I don't have any better ideas, so it seems that we're going to have to live with this now.

EShamaev: Yes, in aviation baro is used a lot but it also changes without "real" connection to any physical altitude. For example when plane is in vicinity of airplane zero baro is set from tower, when in free flight, all baro's zeros on all planes are set to same value.
But this value is really needed to be reported.
For height_agl this one is strictly based on ground and follow terrain but is very good for photo missions to store.

uavcan/equipment/ahrs/1002.PositionAttitude.uavcan
```diff
@@ -0,0 +1,40 @@
...
+#   height (worse)  [meter^2]
+#   roll angle      [radian^2]
+#   pitch angle     [radian^2]
+#   yaw angle       [radian^2]
+#
+float16 pose_covariance[<=36]
+
+float32[3] linear_velocity_ned      # meter/second
+float16[3] angular_velocity         # rad/sec
+float16[3] linear_acceleration      # m/s^2
```
EShamaev: I put it here not for fusion reasons, but as you stated later as a filtered solution to be used right away for logging simple things like uncoordinated turn condition, etc.

pavel-kirienko:
Sorry to keep you hanging Eugene. I suggest to move this message away from the AHRS namespace into a new second level namespace `uavcan.navigation.`. Do you think that makes sense? I am not comfortable with keeping this definition in the equipment namespace because it is not pertaining to any specific kind of equipment.

EShamaev:
Fully agree. Will do.

pavel-kirienko:
Hey Eugene, it is just a reminder that we're one commit away from merging this.

EShamaev:
Moved and added QNH parameter as an essential information on baro altitude calculation

pavel-kirienko:
Superseded by https://github.com/UAVCAN/dsdl/pull/23. I had to open a new pull request because I can't push to the fork owned by @EShamaev.

PR #20, opened 2017-04-12
Added message 1002.MagneticFieldStrength2.uavcan

AlexeyBulatov:
New message based on 1001.MagneticFieldStrength with new field sensor_id, which can be used if your node has more then one sensor.

### DIFF
--- a/uavcan/equipment/ahrs/1002.MagneticFieldStrength2.uavcan
+++ b/uavcan/equipment/ahrs/1002.MagneticFieldStrength2.uavcan
@@ -1,8 +1,9 @@
 #
 # Magnetic field readings, in Gauss, in body frame.
 # SI units are avoided because of float16 range limitations.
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
Thanks Alexey.

Could you please briefly describe the use case that you envision for this message? Why does your application require multiple magnetometers per node? Thanks!

AlexeyBulatov:
The proposed implementation looks like a viable solution in case several sensors need to be used on a device.
If I am not mistaken, another approach (i.e. adding a new node for magnetometer) might make parameter updating more difficult . From simple tests that I conducted it seems like adding another node also yields a bigger binary.

pavel-kirienko:
Thanks @AlexeyBulatov. Seems like nobody have objected to this proposition in 8 days.

I suggest to declare the old message deprecated in the header comment, could you please do that, and I'll merge this ASAP.

Thanks.

AlexeyBulatov:
Thank you @pavel-kirienko!
I have added comment about deprecated message.

PR #23, opened 2017-05-09
Navigation message from Eugene Shamaev

pavel-kirienko:
A couple of minor changes here:

* More elaborate comments.
* The linear velocity, angular velocity, and the linear acceleration are all explicitly marked as belonging to the body frame. This is in accordance with the conventions widely adopted in robotics.

@EShamaev you can now commit directly to this repository.


EShamaev:
LGTM

PR #24, opened 2017-05-27
Fixed description of the velocity covariance matrix

pavel-kirienko:
This change does not affect the layout, only comments are changed.


EShamaev:
Yes, true.

PR #25, opened 2017-05-27
Internet access from UAVCAN networks

pavel-kirienko:
In this pull request, I have outlined my vision of how internet access from UAVCAN networks should be implemented. Please read the comments provided in the data type definitions for explanations.

Feedback is welcome.

## Application examples

* Direct telemetry transmission from UAVCAN nodes with no dependency on external means of broadcasting.
* Creation of web API for onboard equipment. For example, a camera vendor can provide a web API for their products that directly interacts with the cameras on board, bypassing all third party logic.
* Reception of real-time correction data streams (e.g. RTCM RC-104) for precise positioning applications.
* Automatic firmware upgrades directly from vendor's website.

P.S. @EShamaev this is the Internet/LAN access feature we talked about some weeks back. Shall it get merged, you should consider supporting it in your telemetry modems.


uavcan/internet/2500.OutgoingUDPPacket.uavcan
```diff
@@ -0,0 +1,121 @@
...
+# Option flags.
+#
+uint8 FLAG_MASQUERADE = 1       # Expect data back (i.e. instruct the modem to add a new entry to the NAT table)
+uint8 flags
+
+#
+# Domain name or IP address where the payload should be forwarded to.
+#
+Address destination_address
+
```
EShamaev: Did not find clear definition of Destination port in Address.

uavcan/internet/2500.OutgoingUDPPacket.uavcan
```diff
@@ -0,0 +1,121 @@
...
+# the modem nodes about this, so that they could prepare to perform reverse forwarding when the
+# expected data arrives from the remote host. The technique of reverse forwarding is known in
+# networking as IP Masquerading, or (in general) Network Address Translation (NAT). The notification
+# is performed by means of setting one of the corresponding flags defined below.
+#
+# In order to be able to match datagrams received from remote hosts and the local nodes they should
+# be forwarded to, modem nodes are required to keep certain metadata about outgoing datagrams. Such
+# metadata is stored in a data structure referred to as NAT table, where every entry would normally
+# contain at least the following fields:
+#   - The local UDP port number that was used to send the outgoing datagram from.
```
EShamaev: Does local port (Source port) for UDP packet gets chosen by modem?

uavcan/internet/2500.OutgoingUDPPacket.uavcan
```diff
@@ -0,0 +1,121 @@
...
+# Modem nodes are required to be able to accommodate at least 100 entries in the NAT table.
+# The units are seconds.
+#
+uint32 NAT_ENTRY_TTL = 86400
+
+#
+# This field is set to an arbitrary value by the transmitting node in order to be able to
+# match the response with the locally kept context. The function of this field is virtually
+# identical to that of UDP/IP port numbers. Modem nodes do not care about specific values
+# kept in this field.
```
EShamaev: I think they should care and if a new session ID is given with NAT enabled, then a new source port should be used.

uavcan/internet/60.HandleIncomingUDPPacket.uavcan
```diff
@@ -0,0 +1,52 @@
+#
+# This message carries UDP packets sent from a remote host on the Internet or a LAN to a node on the
+# local UAVCAN bus. Please refer to the defintion of the message type OutgoingUDPPacket for a
+# general overview of the packet forwarding logic.
+#
+# This data type has been made a service type rather than a message type in order to make the transfers
+# addressable, allowing the nodes to employ hardware CAN acceptance filters for filtering out
+# forwarded datagrams that are not addressed to them. Additionally, requiring the destination
+# nodes to always respond upon reception of the forwarded datagram opens interesting opportunities
+# for future extensions of the forwarding protocol.
```
EShamaev: This is not required by UDP. What is meant?

uavcan/internet/2500.OutgoingUDPPacket.uavcan
```diff
@@ -0,0 +1,121 @@
...
+#
+# Domain name or IP address where the payload should be forwarded to.
+#
+Address destination_address
+
+#
+# Effective payload. This data will be forwarded to the remote host verbatim.
+# Note that tail array optimization is active here.
+#
+uint8[<=1024] payload
```
EShamaev: Why 1024?

uavcan/internet/60.HandleIncomingUDPPacket.uavcan
```diff
@@ -0,0 +1,52 @@
...
+void8
+
+#
+# Effective payload. This data will be forwarded from the remote host verbatim.
+# If the received datagram contains more data than fits into this field, it should be segmented into
+# multiple consecutive requests. It is not necessary to wait for response of the current request
+# before invoking the next one.
+# Note that tail array optimization is active here.
+#
+uint8[<=1024] payload
```
EShamaev: Why 1024?

uavcan/internet/2500.OutgoingUDPPacket.uavcan
EShamaev: I would prefer limiting it to 508 bytes for compatibility with all hosts.

uavcan/internet/60.HandleIncomingUDPPacket.uavcan
EShamaev: I would prefer limiting it to 508 bytes for compatibility with all hosts.

uavcan/internet/2500.OutgoingUDPPacket.uavcan
pavel-kirienko: Yes. This is a normal practice for UDP masquerading. Also, see this: https://en.wikipedia.org/wiki/UDP_hole_punching

uavcan/internet/2500.OutgoingUDPPacket.uavcan
pavel-kirienko: Indeed, I've removed the do not care clause, it is confusing.

uavcan/internet/2500.OutgoingUDPPacket.uavcan
pavel-kirienko: That was a bug, it is now fixed.

uavcan/internet/60.HandleIncomingUDPPacket.uavcan
pavel-kirienko: The future extensions mentioned in the description would lie outside of the scope of UDP. For example, a local node, having received an expected response from a remote host, could instruct the modem to remove the corresponding masquerading entry by setting a flag in the response data structure. Although this specific use case is rather questionable, in indicates what kind of extensions I'm talking about.

uavcan/internet/2500.OutgoingUDPPacket.uavcan
EShamaev: I know and I think it should be in comments as hole punching is not the only option.

EShamaev: I know and I think it should be clearly stated in comments as hole punching is not the only option.

uavcan/internet/60.HandleIncomingUDPPacket.uavcan
pavel-kirienko: The size of outgoing payload has been limited to 508 bytes per your suggestion, as it makes perfect sense compatibility wise. There is, however, no use limiting the size of incoming payload because we can't guarantee that external hosts will adhere to the 508 bytes limitation. Normally, the maximum size of response payload should be equal to the maximum size of UDP packet, which is 65535 bytes, but data structures this large would be unfeasible for memory constrained, deeply embedded systems that typically run UAVCAN. 1024 bytes seems like an acceptable middle ground; besides, according to [this](https://stackoverflow.com/questions/1098897/what-is-the-largest-safe-udp-packet-size-on-the-internet) and [that](https://stackoverflow.com/questions/14993000/the-most-reliable-and-efficient-udp-packet-size) threads, the vast majority of UDP packets out there are smaller.

EShamaev: Yes, agree. 1024 is still a quite "random" number, maybe changing it to 1432 will make more sense without much burden on memory?

pavel-kirienko: 14**7**2 is marginally less "random" than 1024, yet harder to remember. I'm not sure whether that 0.1% (an uneducated guess) of use cases we're buying with this increase worth the extra 448 bytes of stack memory that would be needed to accommodate a decoded message object. I propose to keep the current maximum unchanged.

pavel-kirienko:
I am debating whether the field `session_id` should be renamed into `local_port` or alike, in order to retain closer similarity with regular UDP.

Interestingly, despite the fact that incoming packets that enter UAVCAN from the Internet do not carry any information about the source address, it is still possible to perform UDP hole punching and establish direct communication channels between UAVCAN nodes that are physically located on different vehicles.

uavcan/internet/60.HandleIncomingUDPPacket.uavcan
EShamaev: I actually meant 1432 with default MTU for Ethernet.
MTU — (Max IP Header Size) — (UDP Header Size) = 1500 — 60 — 8 = 1432

pavel-kirienko: Sorry, you are right. The rest of my argument still stands though.

EShamaev: I really think that 508 is more than enough for UDP. But if you wish, 1024 is OK.

EShamaev:
`session_id` makes more sense that `local_port`
With session ID, packets do not need source address at all.
However there might be a problem with UAVCAN <-> Ehternet <-> UAVCAN communications.
The NAT translation will only happen if there was an outgoing packet. So there should be a possibility to setup 'listening' on specific without having any outgoing packet.

pavel-kirienko:
> The NAT translation will only happen if there was an outgoing packet.

Yes, hence the hole punching technique. It allows hosts to establish a direct communication channel through a single NAT layer on either one or both sides of the link. See https://en.wikipedia.org/wiki/UDP_hole_punching#Flow

However, I don't anticipate high demand for direct links between nodes that belong to different vehicles. Cellular service providers typically do not assign global IP addresses to their subscribers, so communication channels will have to circumvent NATs in provider networks as well, which might be impossible.

EShamaev:
This flow assumes the possibility to define both source and target ports and most important - present of one known public server. With assumption of having white IP server, anything can be done with even simpler flows.
However it might be useful to provide one more flag, that will instruct to use session_id as a source port for modem.

pavel-kirienko:
> This flow assumes the possibility to define both source and target ports.

I don't see why the source port must necessarily be defined. The source port can be chosen arbitrarily by the modem node, and as long as the mapping between this arbitrarily chosen port number **P** and the pair of (UAVCAN node ID, session ID) is known, UDP hole punching should work. Upon reception of any datagram via the port **P** the modem node would simply forward it to the specified node. A downside here is that the targeted node would have no way to know where the datagram has come from, since we provide no source address for incoming datagrams, unless the datagram itself contains some means to identify its origin.

> With assumption of having white IP server, anything can be done with even simpler flows.

Absolutely. This feature is mostly intended for client-server architectures, where clients are nodes and servers are located on the Internet or LAN. The simplest way to arrange communication between nodes is to use an external server as a broker.

EShamaev:
> I don't see why the source port must necessarily be defined.

This is only regarding 2 UAVCAN nodes in different networks. Because 2nd node will have no knowledge on the source port of 1st node. So direct communication will not be possible.
Because of this intermediary server S is needed in flow you provided. It advises to nodes the correct full address of each other.

pavel-kirienko:
Yes, direct communication can only be established with the help of an external server, this is what I implied in my descriptions above. At any rate, we should not focus on this use case too much. It is possible not by design, but rather as a side effect.

nbrakeES:
Opening up the uavcan to the internet/lan communications would bring to mind security concerns.  What security concerns are there if the example applications are achieved?

pavel-kirienko:
Excellent question, Nick. Yes, there are always security concerns when the Internet is involved; however, they are outside of the scope of UAVCAN I think. What we're offering here is merely a gateway between nodes and external hosts, it is up to the nodes themselves to ensure adequate security.

A possible attack vector that comes to mind is as follows: using a network scanning tool, like Nmap, discover the UDP ports opened by the modem node, and flood them with traffic. The modem node would propagate the traffic into the UAVCAN bus. However, I do not expect this attack to have any significant adverse effect on the operation of the bus, because of this clause:

> It is recommended to perform service invocations at the same transfer priority level as was used for broadcasting the latest matching message of type OutgoingUDPPacket. However, meeting this recommendation would require the modem node to implement additional logic, which may be undesirable. Therefore, implementers are free to deviate from this recommendation and resort to a fixed priority level instead. In the case of a fixed priority level, lower priorities should be preferred in order to avoid interference with real-time traffic on the bus. It is advised to use priority level values from 31 to 24, inclusive (remember that numerically higher values correspond to lower priority levels).

It guarantees that the traffic generated by such flooding attack will not affect high-priority communications on the bus.

Please post here if other attack vectors come to mind.

pavel-kirienko:
I intend to have this merged in a week or so unless new change propositions are brought up.

pavel-kirienko:
The definitions have been updated. I have incorporated the ideas we came up with during the conversation on the mailing list with @olliw42. 

The new data types allow the use of transport protocols other than UDP.

uxvrob:
Here are just some thoughts I had when considering the DroneCode ecosystem and UAVCAN/dsdl

I have read about the CoAP and it may be a useful implementation somewhere in the ecosystem. It's designed to supervise or control remotely IoT type devices ie. sensors, acutators etc.  The interesting thing is that it runs over UDP packets with an optional DTLS encryption with generally low overhead, multicast, and has a REST model.

Particle.io probably has the best implementation of CoAP, DTLS encryption and web-authentication that I've seen (and its open source!!)

- https://github.com/spark/firmware/blob/develop/communication/README.md

As QGC and Pixhawk prefer Qt as the default build environment and IDE,  Qt 5.9 has a serial bus framework for CAN based protocols.  I envision the UAVCAN implementation in QGC may be coupled to this framework in the future:

- https://github.com/qt/qtserialbus

Qt is also working on a CoAP implementation:

- https://wiki.qt.io/QtCS_2017_CoAP

This could be added to QGC once it is developed.  

Now, if I was building a web application to receive UAVCAN data, I would only want to receive it from vehicles that are registered and setup through my application.  

I would like to assign them a UUID and a private encryption key for authentication.  

I think there should be UAVCAN data fields for these with the UUID stored on every node and a PGP key stored on the communication bridge device.  

On system startup, a service would sync with the cloud server through authentication handshakes and a web micro service would be started to handle data transmission.

Considering CoAP, here are some basic data streaming situations:  

- MAVLINK -> UAVCAN -> CoAP -> Web Service
- UAVCAN -> CoAP -> Web Service
- UAVCAN -> GCS -> CoAP -> Web Service

Another item to consider, there are soooo many GCS systems (web & ground based) or flight data collection software being developed, there would need to be an OpenAuth type of standardization for IoT devices or UV systems in general.  

Hopefully these are useful points.

pavel-kirienko:
Thanks, Robbie, this is useful. CoAP seems like an interesting protocol, I never heard of it before. Do you think the UDP transport defined in this PR is sufficient for a full-featured implementation of a CoAP client? If not, what is missing?

uxvrob:
Haha not sure Pavel.  @EShamaev and yourself, to me, are more knowledgeable to answer.  Although, implementing CoAP with DTLS can also resolve the security concerns brought by @nbrakeES.

Now, if the UDP transport in this PR IS sufficient for a full-featured CoAP client, what issues could there be at the hardware level?  What issues will there be in industry adoption?

We know it will add more program space and processing overhead, although consider the implications in 5 years where unmanned systems have these protocol layers embedded and they are impossible to change.    

Some manufacturers may not implement this because it may not meet a IETF (web protocol), IEEE (electrical & hardware design) and/or ISO (manufacturing quality) standard.  That needs to be addressed.

Sorry, I may have gone off on a tangent to the topic. Although since it's a public post, I hope people reading will contribute their creative thoughts and solutions to developing this technology further.

Big thanks to @pavel-kirienko for spearheading UAVCAN and the many contributors here pushing it forward.

Here are some reference docs:

[IETF - RFC 7252 - Official CoAP specification](http://tools.ietf.org/html/rfc7252)

[ARM CoAP Tutorial April 30 2014.pdf](https://github.com/UAVCAN/dsdl/files/1405411/ARM.CoAP.Tutorial.April.30.2014.pdf)

[coap-cheatsheet.pdf](https://github.com/UAVCAN/dsdl/files/1405406/coap-cheatsheet.pdf)

pavel-kirienko:
Reimplemented in #47, closing.

PR #26, opened 2017-06-07
Add raw airspeed differential press and temp message

bugobliterator:
This PR adds Raw Airspeed Data message for use in transmitting the differential pressure and temperature directly to Autopilot or any other consumer where it will be converted into True Airspeed.

### DIFF
--- a/uavcan/equipment/air_data/1027.RawAirData.uavcan
+++ b/uavcan/equipment/air_data/1027.RawAirData.uavcan
@@ -1,0 +1,41 @@
+#
+# Raw Air Data.
+#
+
+# Note: unused vars should be assigned NaN
+
+#
+# Heater State
+# 
+uint8 FLAG_HEATER_AVAILABLE      = 1
+uint8 FLAG_HEATER_WORKING        = 2
+uint8 FLAG_HEATER_OVERCURRENT    = 4
+uint8 FLAG_HEATER_OPENCIRCUIT    = 8
+uint8 flags
+
+#
+# Pressure Data
+#
+float32 static_pressure                 # Pascal
+float32 differential_pressure           # Pascal
+
+#
+# Temperature Data
+#
+float16 static_pressure_sensor_temperature          # Kelvin
+float16 differential_pressure_sensor_temperature    # Kelvin
+
+float16 static_air_temperature          # Kelvin
+                                        # This field contains the raw temperature reading 
+                                        # from the externally mounted temperature sensor or, 
+                                        # in absence of one, the raw temperature of the pressure sensor.
+
+float16 pitot_temperature               # Kelvin
+
+
+float16[<=16] covariance                # order of diagonal elements : 
+                                        # static_pressure, differential_pressure,
+                                        # static_air_temperature, pitot_temperature
+                                        # Pascal^2 for pressure variance and covariance
+                                        # Kevin^2 for Temperature variance and covariance
+                                        # Pascal*Kelvin for pressure/temperature covariance


pavel-kirienko:
Please align the comments and add a new line in the end of the message definition.

Calling in @EShamaev. Does this make sense to you? Any comments?

bugobliterator:
@pavel-kirienko I have fixed the issues.

pavel-kirienko:
Thank you. We're waiting for Eugene to approve/reject it now.

EShamaev:
I would also advise to have static pressure and pitot temperature reported.

pavel-kirienko:
Makes sense. The following then should be explicitly specified in the comments then:

* What exactly is understood under the plain `temperature` field. Perhaps it should be renamed.
* How do we represent the unknown quantities, and which ones are required. Normally we use NaN to indicate that a certain quantity is unspecified.

EShamaev:
Temperature should be OAT. Second one can be pitot_temperature.
If it is not available - NaN is OK.

pavel-kirienko:
Okay. Do we need error variance estimates for all of the reported parameters?

EShamaev:
If we have 4 params, then matrix will be good as they are related to each other.

pavel-kirienko:
OK, thanks for your inputs Eugene.

EShamaev:
Some flags would be nice as well. Most essential: HEATER_WORKING, HEATER_OVERCURRENT, HEATER_OPENCIRCUIT.

EShamaev:
For reference, here is a list of ARINC429 labels usually sent from airdata computers:
image

bugobliterator:
@EShamaev can you specifically list out the additional messages I should add. For my purposes, I am using a simple module based on MS4525 which only reports Diff Pressure and Static Air Temperature, and no additional processing is being done before sending that data over UAVCAN. If you feel additional params should be listed, I would gladly add them to this PR. If you feel like adding them, I am fine with that as well.

EShamaev:
When message is added to DSDL there will be no chance to change it in future. So we need to carefully think what is needed for broad range of devices that might be using it.

I propose to have following:

static_pressure [Pascal]
differential_pressure [Pascal]
OAT [K]
pitot_temperature [K]
variances [4x4]
flags [HEATER_WORKING, HEATER_OVERCURRENT, HEATER_OPENCIRCUIT]

Also correct naming should be Raw Air Data. I also have the same device developed and was going to post a PR regarding addition of such packet, but I have real external OAT sensor, pitot heater with temp and current sensors and static pressure sensor.
Basicly the proposed list is all data that is coming from a real world (big planes) pitot sensors / airdata sensors.
The pitot temperature and heater state is extremely important for determining the icing on pitot.

Also regarding MS4525 you mentioned, it is correct to say that it is reporting the sensor's temperature and not the "Static Air Temperature", however for such use case we consider this as OAT.

LorenzMeier:
We need two different temperatures for static and dynamic pressure. I tend to find that the temperature compensation from the sensors doesn't work and we need the full data in the autopilot in order to post-compensate.

EShamaev:
Then please add 2 separate temperatures. Sensor temperature is **not** OAT. If a device do not have OAT sensor it may broadcast sensor's temperature as OAT but there are devices with external OAT sensor.

bugobliterator:
ok will do

EShamaev:
>> We need two different temperatures for static and dynamic pressure. I tend to find that the temperature compensation from the sensors doesn't work and we need the full data in the autopilot in order to post-compensate.

The sensor that is sending this message is "smart" and has it's own MCU. I am thinking now that this compensation should be in the sensor, not in AP.
Would like to hear other opinions on this.

bugobliterator:
@EShamaev There could be cases where compensation is not implemented for any reason, but AP does have Temperature based calibration feature like in the case of PX4, where user can calibrate wrt temperature on there end.

EShamaev:
Alright, these 2 temperatures should be in message then, but I still think that calibration should be done on the sensor. (For newly developed devices)

bugobliterator:
@EShamaev @pavel-kirienko I have made the suggested changes. Please review and do the needful.

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,29 @@
...
+#
+# Pressure Data
+#
+float32 static_pressure                 # Pascal
+float32 differential_pressure           # Pascal
+
+#
+# Temperature Data
+#
+float16 sensor_temperature              # Kelvin
```
EShamaev: There should be 2 sensor temperatures

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,29 @@
...
+float16 static_air_temperature          # Kelvin
+float16 pitot_temperature               # Kelvin
+
+float16[<=25] covariance                # Pascal^2 for pressure variance and covariance
```
EShamaev: I feel that sensor temperatures do not need covariance

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,29 @@
...
+# Pressure Data
+#
+float32 static_pressure                 # Pascal
+float32 differential_pressure           # Pascal
+
+#
+# Temperature Data
+#
+float16 sensor_temperature              # Kelvin
+float16 static_air_temperature          # Kelvin
```
EShamaev: Just came to mind that Static air temperature is actually not the raw temperature read by sensor, despite it being called OAT sensor.
This is not really huge matter for regular UAV's low speeds, and can be left as it is with addition of an appropriate comment.

bugobliterator: what comment do you advise to be put here? will this do : `this field should contain the RAW temperature reading from the sensor needed for actual static_air_temperature calculation`

EShamaev: Something like:

This field contains the raw temperature reading from the externally mounted temperature sensor or, in absence of one, the raw temperature of the pressure sensor.

uavcan/equipment/air_data/1027.RawAirData.uavcan
bugobliterator: do you mean separate temperatures for static and differential pressure sensors?

EShamaev: As far as I understood yes, as there might be 2 separate sensors. 
However I am still in doubt about this. 
I don't think that anyone ever used precise calibration equipment to check the calibration curves. I expect that the need for such calibration on AP side is coming from wrong reading of outside temperature (without dedicated sensor) and thus the wrong conversion to TAS.
In both cases the issue is present and post calibration on AP side can deal with that to some extent.

WickedShell:
I know I don't normally comment on these things, and I could be missing the definition, but what are the expected flags for no data? NaN's or a negative number? I know some of the other files have comments that say unknown fields should be NaN but I'm not seeing that in this one?

pavel-kirienko:
> I know some of the other files have comments that say unknown fields should be NaN but I'm not seeing that in this one?

It's a common convention in UAVCAN (it is mentioned in the specification) to use NaN as a placeholder for unspecified values. It should be stated explicitly in the definition here though in order to avoid confusion. @bugobliterator could you please add that? Thanks!

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,29 @@
...
+#
+
+#
+# Heater State
+# 
+uint8 HEATER_UNAVAILABLE    = 0
+uint8 HEATER_WORKING        = 1
+uint8 HEATER_OVERCURRENT    = 2
+uint8 HEATER_OPENCIRCUIT    = 3
+uint8 heater_state
```
pavel-kirienko: Please rename `heater_state` to `flags` in order to facilitate future extensibility of the data type. We may want to add new flags in the future that will have nothing to do with the heater. The constants should be renamed accordingly, e.g. `FLAG_HEATER_WORKING`, etc.

Concerning the field `FLAG_HEATER_UNAVAILABLE`, can we please avoid negative logic here, it's slightly more confusing than the reverse. I suggest renaming into `FLAG_HEATER_AVAILABLE`.

uavcan/equipment/air_data/1027.RawAirData.uavcan
pavel-kirienko: @EShamaev Could you please list what variances we need here? Given the uncertain nature of OAT measurements, it might be useful to provide variance estimates at least for that?

@bugobliterator The ordering of the diagonal elements in the covariance matrix should be specified explicitly. Typically it would match the ordering of the preceding fields.

uavcan/equipment/air_data/1027.RawAirData.uavcan
pavel-kirienko: Should we clarify here that we're talking about a static air temperature sensor rather than just a plain temperature sensor, since OAT sensors may incorporate corrections for the distortions caused by the flow of air?

> This field contains the raw temperature reading from the externally mounted static air temperature sensor, or, in the absence of one, the raw temperature of the pressure sensor.

bugobliterator:
@pavel-kirienko @EShamaev made the suggested change. Please review.

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,41 @@
+#
+# Raw Air Data.
+#
+
+# Note: unused vars should be assigned NaN
```
pavel-kirienko: Splendid. Which ones are required though?

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,41 @@
...
+# Pressure Data
+#
+float32 static_pressure                 # Pascal
+float32 differential_pressure           # Pascal
+
+#
+# Temperature Data
+#
+float16 baro_sensor_temperature         # Kelvin
+float16 aspd_sensor_temperature         # Kelvin
```
pavel-kirienko: Does ASPD mean "airspeed pressure difference"? Perhaps we should spell it out, I don't think this acronym is widely known?

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,41 @@
...
+uint8 FLAG_HEATER_WORKING        = 1
+uint8 FLAG_HEATER_OVERCURRENT    = 2
+uint8 FLAG_HEATER_OPENCIRCUIT    = 3
+uint8 flags
+
+#
+# Pressure Data
+#
+float32 static_pressure                 # Pascal
+float32 differential_pressure           # Pascal
```
pavel-kirienko: Perhaps we could make do with float16 for the differential pressure measurement? What range and resolution are typically required?

uavcan/equipment/air_data/1027.RawAirData.uavcan
bugobliterator: That should be left to the implementer. I think if you see all the params each one of them is optional. One can choose to send either baro related raw info or airspeed related or both.

uavcan/equipment/air_data/1027.RawAirData.uavcan
bugobliterator: I don't think that would be a good idea, the value can range from ~1 to 300Pa (considering max airspeed of ~25m/s, it can go higher with faster planes). we should atleast give way for 0.01 resolution in the whole range, which I doubt will happen with float16.

uavcan/equipment/air_data/1027.RawAirData.uavcan
bugobliterator: I will change this to airspd_sensor_temperature. Will that work?

EShamaev: I think that static_sensor_temperature and differential_sensor_temperature is the most precise.

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,41 @@
...
+# Note: unused vars should be assigned NaN
+
+#
+# Heater State
+# 
+uint8 FLAG_HEATER_AVAILABLE      = 0
+uint8 FLAG_HEATER_WORKING        = 1
+uint8 FLAG_HEATER_OVERCURRENT    = 2
+uint8 FLAG_HEATER_OPENCIRCUIT    = 3
+uint8 flags
```
EShamaev: Maybe
FLAG_HEATER_AVAILABLE = 1
FLAG_HEATER_WORKING = 2
FLAG_HEATER_OVERCURRENT = 4
FLAG_HEATER_OPENCIRCUIT = 8
?
It will be more flag-like

pavel-kirienko: Hah, good catch, I entirely overlooked it. Of course, the current values won't work.

uavcan/equipment/air_data/1027.RawAirData.uavcan
pavel-kirienko: Good point @bugobliterator.

@EShamaev do we expect that raw air measurements will be typically produced by the same physical device? If that is not entirely likely, perhaps we should consider separation of this message into two smaller ones, one for the barometric data and the other for the differential pressure data? Although that would deprive us of the covariance matrix.

bugobliterator: @pavel-kirienko I don't see such data coming from the same sensor. But, same module sending both is entirely possible. My initial PR was for only one kind of raw data i.e. AirSpeed related, as my module only sends that data. I do like the idea of splitting them up. I do see a possibility of correlation b/w static and diff pressure. Although, not sure if that's going to be used or even be needed.

pavel-kirienko: Thanks, what I implied is the same device, yes, not the same sensor. We should wait for Mr. Shamaev's comments on this.

WickedShell: Merits to either approach, 2 messages definetly fits better with the cheap sensors we are all using at the moment, a combined message makes more sense on nicer/higher end equipment in the future. But given that the first use case will probably be cheap sensors that don't include both parts, and that there are already static pressure/temp messages it might make more sense to keep them split up...

EShamaev: Definitely it can come from a single device.
And usually is, the use of cheap sensors and separation of data is to my opinion an exclusion from usual practice.

bugobliterator:
@pavel-kirienko @EShamaev Made the fixes. Can you review...

uavcan/equipment/air_data/1027.RawAirData.uavcan
pavel-kirienko: Ok, let us keep it in a single message then.

uavcan/equipment/air_data/1027.RawAirData.uavcan
```diff
@@ -0,0 +1,41 @@
...
+# Pressure Data
+#
+float32 static_pressure                 # Pascal
+float32 differential_pressure           # Pascal
+
+#
+# Temperature Data
+#
+float16 static_p_sensor_temperature     # Kelvin
+float16 diff_p_sensor_temperature       # Kelvin
```
pavel-kirienko: Not to be nitpicking, but can we not say `static_pressure_sensor_temperature` and `differential_pressure_sensor_temperature`? I promise that won't make the code run slower or anything.

bugobliterator: @pavel-kirienko Just feel that its too long var name and I don't see any additional info being given by increasing it to that length.

pavel-kirienko: I'm not entirely convinced of the advantages of shorter names. They make the code slightly less intelligible, so the policy should be to avoid uncommon abbreviations whenever possible. The additional info given by using a longer name is that it becomes clear that we're talking about a pressure sensor here, and not about a P-sensor, which could mean, say, a proximity sensor. Granted, the reader will be able to understand what does the P mean in this context, but it is better to spare them the extra mental effort whenever possible.

bugobliterator:
@pavel-kirienko made the changes to variable.

EShamaev:
I am happy with this

pavel-kirienko:
Thank you, everyone.

PR #27, opened 2017-06-08
Siddharth raw IMU message

pavel-kirienko:
# RawIMU

@bugobliterator I have reworked the raw IMU message and now it is ready to be merged. The following changes have been introduced:

* The data type of the integrals has been upgraded to float32 because float16 does not provide sufficient resolution for many use cases.
* The timestamp and the integration interval fields have been made common because rate gyros and accelerometers are typically sampled synchronously or nearly so. Also, having separate timestamp and interval fields makes the message unfit for single-frame transfers over CAN FD, which is a major problem for single-frame messages like this one.
* Axis ordering and units were clarified.

Please let me know if the changes make sense to you. If they do, I will merge it right away.

# ControlInput

Unfortunately, I cannot accept the message ControlInput right now, because it is not sufficiently well-defined. The proposed definition has at least the following shortcomings:

* The purpose and the semantics of the field `controller_id` is not defined.
* The proposed set of units does not make sense (see the field `control_type` and the related commands). For example, what use is there for a message that carries an array of control inputs in the meter units?
Why is the attitude represented in radian/second?
* Why is there only 8 control inputs?

You should go back to the whiteboard and carefully think about the use cases that your definition should cover.

If you would like to hear my take on this, then I would propose this. First of all, what we're defining here seems to be an abstract real-time manual control input from a human operator. The operator would probably have a number of scalar control inputs, that will need mapping onto functions on different UAVCAN nodes. I am not sure whether we should explicitly define the units of measurements for the control inputs since we don't know the semantics of their use, therefore I propose to resort to abstract unitless quantities. The resulting solution then would be described in 2 messages (the names may need to be changed) roughly like this:

### `uavcan.equipment.manual_control.Channel`

```python
uint8 channel_id
uint16 value            # where 0 corresponds to the minimum, 65535 to the maximum
```

### `uavcan.equipment.manual_control.Input`

```
Channel[<=18] channels
```


bugobliterator:
@pavel-kirienko Agreed with your comments. I will review the ControlInput part, will try to come up with something that is more agreeable.

pavel-kirienko:
Thank you. Shall we merge this now?

bugobliterator:
yes please. Thanks

Issue #18, opened 2017-04-02
Configure Travis to check correctness of the definitions


pavel-kirienko:
On several occasions faulty changes have been submitted to master. A CI setup would have allowed to weed out such issues.

PR #28, opened 2017-06-15
Fix duplicate variable ids

bugobliterator:
@pavel-kirienko Sorry for breaking master! Should have checked that before requesting to to pull my PR.


pavel-kirienko:
Thanks for fixing this. It's a serious problem that we don't check this automatically. I should implement this check soon.

bugobliterator:
I will try and add a travis check for this. So that this can be avoided.

pavel-kirienko:
Awesome.

bugobliterator:
I think simply running dsdl compiler on it should do. Do you agree.

pavel-kirienko:
Yes, it should be enough. We have a ticket for this: https://github.com/UAVCAN/dsdl/issues/18

PR #31, opened 2017-06-24
Electronic fuel injection status message

pavel-kirienko:
This is a prototype that we are working on in cooperation with Ecotrons people. Please review this with the utmost scrutiny.

FYI @yyl123hf @EShamaev @frizensami @oxinarf


pavel-kirienko:
See https://github.com/ArduPilot/ardupilot/pull/6438 for context.

uavcan/equipment/ecu/1120.Status.uavcan
```diff
@@ -0,0 +1,94 @@
...
+# The estimated value should be low-pass filtered in order to prevent aliasing effects.
+# Units: (centimeter^3)/minute.
+#
+float32 fuel_consumption_rate_cm3pm
+
+#
+# Estimate of the remaining fuel.
+# Units: centimeter^3.
+#
+float32 remaining_fuel_volume_cm3
```
EShamaev: Not sure if it needs to be here?
There are cases when there are more than one fuel tank and also a lot of different sensors.

uavcan/equipment/ecu/1120.Status.uavcan
```diff
@@ -0,0 +1,94 @@
...
+
+#
+# Estimate of the remaining fuel.
+# Units: percent.
+#
+uint7 remaining_fuel_percent
```
EShamaev: Same as above

pavel-kirienko:
Could you propose a replacement message please, Eugene?

EShamaev:
I propose to add:

into 1120:
- exhaust gas temperature (EGT) (preferably more than one) (this is a must!)
- oil pressure
- oil temperature
- fuel pressure

into CylinderStatus:
- cylinder head temp (CHT)

If we are to support turbos, then additionally:
- NG (Percent or RPM)
- NF (Percent or RPM)
- NP (RPM)
- Torque (Nm)
- T4 (Kelvin)

EShamaev:
> Could you propose a replacement message please, Eugene?

I think that message for reporting readings from fuel level sensors would do the trick :)

uavcan/equipment/ecu/1120.Status.uavcan
EShamaev: I think the idea was to have "calculated"  remaining fuel. If that's the case - better would be to have "calculated" consumed volume.

uavcan/equipment/ecu/1120.Status.uavcan
```diff
@@ -0,0 +1,118 @@
...
+# Engine coolant temperature.
+# Units: kelvin.
+#
+float16 coolant_temperature
+
+#
+# Exhaust gas temperature (EGT).
+# Units: kelvin.
+#
+float16[<=16] exhaust_gas_temperature
```
pavel-kirienko: I suppose each value corresponds to a particular cylinder? In that case, this parameter should be moved into the CylinderStatus struct.

uavcan/equipment/ecu/1120.Status.uavcan
pavel-kirienko: This makes sense.

uavcan/equipment/ecu/CylinderStatus.uavcan
```diff
@@ -0,0 +1,22 @@
...
+# Fuel injection time.
+# Units: millisecond.
+#
+float16 injection_time_ms
+
+#
+# Cylinder head temp (CHT).
+# Units: kelvin.
+#
+float16 cylinder_head_temp
```
pavel-kirienko: Surely you meant `temperature`, not `temp`. ;)

uavcan/equipment/ecu/1120.Status.uavcan
EShamaev: Not necessarily.
Number of exhaust gas temperatures is usually less or equal to the number of cylinders.
Depends on engine construction.

frizensami:
It's very likely that certain engine control systems will not provide all of this data, in this particular case, the Ecotrons EFI doesn't provide CHT/oil temp. Is there a standard way of dealing with a UAVCAN message that may not have all its fields filled?

EShamaev:
Usually NaN is transmitted when the parameter is not available.

frizensami:
Okay, understood. @pavel-kirienko, have Ecotrons begun to use this message definition?

uavcan/equipment/ecu/1120.Status.uavcan
pavel-kirienko: > Number of exhaust gas temperatures is usually less or equal to the number of cylinders.

In that case, it would make sense to add an optional field to the cylinder status message?

pavel-kirienko:
I suggest we rename the namespace from `ecu` to `piston_ecu`, for the sake of future extensibility to other types of engines (e.g. turbos that Eugene mentioned above).

@frizensami Not yet.

pavel-kirienko:
Okay. I took the liberty to relocate the exhaust gas temperature field into the cylinder status message. @EShamaev please review.

The remaining fuel estimates were removed in favor of a dedicated message, that should be located in a separate namespace in order to facilitate its reuse for other types of engines.

EShamaev:
Yes, agree.

pavel-kirienko:
What data do we need to carry in the fuel status message? I suppose at least the following fields are required:

* Fuel tank ID
* Remaining fuel, cm<sup>3</sup>
* Consumed fuel, cm<sup>3</sup>
* Fuel consumption rate, cm<sup>3</sup>/min
* Fuel temperature, kelvin

EShamaev:
The fuel can not only be consumed, but also transferred between tanks.
Do not have a clear idea right now how to implement this in messages, need to think about it.

pavel-kirienko:
> Do not have a clear idea right now how to implement this in messages, need to think about it.

Negative consumption rate?

EShamaev:
What bothers me is that this action does not originate from tank fuel level sensor, but from other equipment.
If we refer to real airplanes, then fuel consumption rate is actually a separate set of flow sensors and has not connection to tank.

pavel-kirienko:
OK, we're going to stand by for your suggestions for a while.

EShamaev:
As we agreed that fuel status message should be separate, it does not delay this one.

srichard95:
Hello,

I have a few idea about the ECU message fields. Does it make sense to add a field which determine the current status of the engine?

#
# Engine status
#
uint8 ENGINE_STOP                     = 1
uint8 ENGINE_SELFSTARTING      = 2
uint8 ENGINE_RUNNING             = 4
uint8 ENGINE_STOP_ERROR        = 8
uint8 engine_status

The oxygen sensor value also an important information and also can be added, but it depends on what is the purpose of the message itself.

As for the fuel status messages. Maybe you can create a message which gives information about the fuel globally and create a separate one for each individual tank. 

Fuel status message:

- Remaining fuel, cm3
- Consumed fuel, cm3
- Fuel consumption rate, cm3/min
- Fuel temperature, kelvin

Fuel tank status message:

- Fuel tank ID
- Transferred quantity [+/-]
- Fuel level [%]
- Fuel temperature

Have a nice day!

pavel-kirienko:
Thanks, Richard,

I don't anticipate much use for the oxygen sensor information outside of the engine control system, but I'd be happy to be corrected on this.

Concerning the fuel status, what use case is there for a global fuel status message? This information can always be derived by consolidating the data from all per-tank fuel status messages.

pavel-kirienko:
> uint8 ENGINE_STOP = 1
> uint8 ENGINE_SELFSTARTING = 2
> uint8 ENGINE_RUNNING = 4
> uint8 ENGINE_STOP_ERROR = 8
> uint8 engine_status

@srichard95 could you please elaborate the meaning for each of the flags you proposed here? They should be joined with the existing flags field.

frizensami:
@pavel-kirienko our team uses information from the oxygen sensor to determine the richness or leanness of the fuel mixture during tuning. I believe it would be helpful to transmit this data elsewhere to be logged or analyzed or acted upon.

EShamaev:
While oxygen sensor seems reasonable, I have seen it only few times in real applications in aviation.
Usually EGT is used to determine fuel mixture and is one of the main values.
The value can be added for generality though.

srichard95:
image

@pavel-kirienko It is based on a state machine. This state machine allow us to track the operation state of the engine. STOP state is present when the engine is not active. Selfstarting state is when you spinning up the engine to start it. Running state is present when the engine is running and ready for load it. STOP_ERROR is set when an unexpected event happened for example run out of fuel, fuel pressure malfunciton, emergency shutdown for overheat etc.

As for my earlier experiances, this structure helps to determine the different operational states. You can make conditions easily using this state machine.

frizensami:
@srichard95 I think tracking the current engine state would be very useful. I think the OVERHEATING and CRANK_ERROR flags should be moved into an `error_flags` variable perhaps, whereas the current engine state as in your state machine should be in a `engine_state_flags` variable. 

I'm not sure if all engines will have this particular operational RPM range though, so I would not recommend adding the additional dependency of checking the RPM range (if that is what you are suggesting).

EShamaev:
Regarding oxygen sensor, what will be the preferred units of measure?
ppm, ppt, lambda, %, mV or smth else?

frizensami:
As far as I know, the Ecotrons ALM-CAN (http://www.ecotrons.com/products/accurate_lambda_meter/) outputs O2 %, that's my only reference point.

yyl123hf:
What does FLAG_OVERHEATING mean?

srichard95:
@EShamaev We usually use the lambda unit, but i think it's all the same.
@yyl123hf FLAG_OVERHEATING can be set when the engine reaches the absolute maximum operational temperature (if i understand it correctly). for ex. above 100°c the bit is set.

yyl123hf:
Engine coolant temperature Can this variable contain the meaning you want to express?

srichard95:
@yyl123hf Yes and no, my opinion is the engine coolant temperature is give an information about the engine temperature itself. There are many types of engine and they have different absolute maximum temperature ratings. The FLAG_OVERHEATING is set, when the engine reach that temperature which is programmed into the ECU according to the engine parameters. That means the other controllers which are read the message of the ECU can detect when the engine is overheated and this information is provided by the ECU and not calculated from the engine coolant temperature on the receiver side. Does it make sense for you?

EShamaev:
Actually CHT is the right value to use for overheat detection.

frizensami:
We should decide if we want the EFI manufacturer to be responsible for telling us when the engine is considered to be overheating. 

If this is an EFI system, perhaps it does not know what this temperature limit is, since it could be connected to any type of engine, and the EFI likely is not responsible for maintaining the engine's temperature as opposed to the end-user providing adequate cooling during operation. 

A counterargument is that having the flag there gives manufacturers a choice if they want to provide this overheating info. A counter-counter argument is that leaving the flag there makes users expect that the EFI will provide such information, even if it does not. I am not sure how to indicate to the user that a flag may be meaningless, whereas normal values can be set to NaN to indicate that.

EShamaev:
> We should decide if we want the EFI manufacturer to be responsible for telling us when the engine is considered to be overheating.
Definitively that is something specific to engine, so limits must be configured in ECU. If we want to run additional checks - we have the temperature reported as well.

frizensami:
@EShamaev my concern is that for non-flag variables, we can set them to NaN to indicate that the EFI does not provide that field. I do not believe we can do this for bits that are in a flag variable, since that flag bit can only be 0 or 1. 

This inability to indicate whether a flag bit is used or not means one of three things (that I can think of):
1. The message will not be self-contained: manufacturers must indicate which flag bits are being used in their own manuals, and users must write their application logic appropriately, or
2. If we require the message to be self-contained, this means we are saying that manufacturers _must_ truthfully support FLAG_STARTING/FLAG_CRANKSHAFT_ERROR/FLAG_OVERHEATING, or
3. When the user sees that (for example) FLAG_CRANKSHAFT_ERROR = 0, the user cannot assume that there really is no such error, just that perhaps the EFI does not provide this data.

Does this make sense? Sorry for being pedantic, just that I expect a lot of deviation in the set of data that EFI manufacturers will provide, and I'm hoping this status message can be unambiguous in spite of that.

EShamaev:
I can propose to have one more variable to be sent (the same width as flags) that will hold 1 in appropriate places if the ECU is supplying the flag.

frizensami:
That sounds perfect. It'll add some overhead, but I think it will be worth it.

frizensami:
Is it standard that all temperatures are in Kelvin? Other units are not in SI, such as fuel_consumption_rate_cm3pm, I suppose for ease of use. Is this to avoid internationalization problems with Celsius and Fahrenheit?

pavel-kirienko:
> For now we should get reply from Ecotrons and maybe Currawong Eng?

I have contacted them via email and also posted an announcement on the mailing list. That should get us more feedback.

> our team uses information from the oxygen sensor to determine the richness or leanness of the fuel mixture during tuning. I believe it would be helpful to transmit this data elsewhere to be logged or analyzed or acted upon.

> As far as I know, the Ecotrons ALM-CAN (http://www.ecotrons.com/products/accurate_lambda_meter/) outputs O2 %, that's my only reference point.

@frizensami @yyl123hf Is that weight percentage or volumetric percentage? Are there any other conventions?

> Is it standard that all temperatures are in Kelvin? Other units are not in SI, such as fuel_consumption_rate_cm3pm, I suppose for ease of use. Is this to avoid internationalization problems with Celsius and Fahrenheit?

We use SI everywhere unless it impairs performance. In the case of the fuel consumption rate estimate, use of the SI unit, which is meter<sup>3</sup>/second, would require a larger data type (float32).

EShamaev:
>  Are there any other conventions?

I think either lambda or ppt will be most understandable.

frizensami:
> @frizensami @yyl123hf Is that weight percentage or volumetric percentage? Are there any other conventions?`

The manual states "concentration" when it talks about O2%, so I believe that is volume. 

Apologies, I was just taking a closer look at the manual. Lambda is also provided by the sensor. I'm using lambda on my end as well to do tuning. I think @EShamaev is right here, lambda will be the most understandable, let's stick with that.

> We use SI everywhere unless it impairs performance. In the case of the fuel consumption rate estimate, use of the SI unit, which is meter3/second, would require a larger data type (float32).

Okay that's very clear, thank you.

pavel-kirienko:
Please review my latest changes. I attempted to incorporate all of the points we discussed earlier; let me know if anything is amiss.

I also add a new FuelTankStatus message that contains the essential data. The integrated quantities, such as consumed fuel and the amount of transferred fuel, were removed because they create a serious ambiguity concerning the point of time where the integration starts from. If you believe this definition is suboptimal, we can freely remove it for now.

frizensami:
Looks great, Pavel. I've asked the mechanical engineers from the team I'm working with (Yonah) to have a look at the current message as well and see if we might have something else to add.

yyl123hf:
Why the unit of fuel consumption is chosen cubic centimeters per minute instead of grams per minute.  The parameters provided by the injector manufacturer are grams per minute as a calculation reference. Getting a weight unit should be easier.

frizensami:
@yyl123hf  Are most injectors based on grams per minute?

yyl123hf:
Most of the things we've used are grams per minute. We generally use Bosch's injector.

frizensami:
https://en.wikipedia.org/wiki/Mass_flow_rate tells me that the SI derived unit for mass flow rate is kg/s, so perhaps using g/min is still alright? It depends on whether we consider fuel flow rate in terms of SI volumetric flow rate (m^3 / s) or SI mass flow rate (kg / s). I suppose if there are no objections to using g/min, it would be easier to integrate with existing systems like Ecotrons'.

frizensami:
Okay, there are no more comments from my team. Besides the discussion on the units for fuel flow rate, I think the rest of the message is good :+1:

pavel-kirienko:
Okay. So it seems that volumetric flow rate should be preferred over the mass flow rate due to the conventions used in the industry. CANaerospace & ARINC825, for example, use liters per hour, which is a volumetric quantity. Ecotrons could easily convert from mass flow rate to volumetric flow rate if they had a fuel density model.

@EShamaev please advice.

EShamaev:
Actually, most ECU's are not measuring the fuel flow at all. At least very few do. Instead it is computed from injection time and pressure. Even if flow is measured, the volume is measured, not mass. So definitively it is volume at first place.
But this goes in contradiction to fuel consumption, which is usually expressed in mass. That is because the engines consume defined mass of fuel to produce defined power.
This issue is much more complex that may appear at first. In ARINC-429 flow is transmitted in volume, but also a separate label with fuel density exist.

My guess is that we go with volume as it is much more natural to most ECUs and FADECs, but we also have a fuel temperature. So calculations into mass might be done after if the ECU supplies us with both parameters.

frizensami:
Okay, if Ecotrons is agreeable with volumetric units, then that will be
fine. There was a mention above that the bosch fuel injector uses mass,
though. @yyl123hf, is using cm^3 okay? Does the ECU model fuel density?


On 28 Jun 2017 00:47, "Eugene Shamaev" <notifications@github.com> wrote:

Actually, most ECU's are not measuring the fuel flow at all. At least very
few do. Instead it is computed from injection time and pressure. Even if
flow is measured, the volume is measured, not mass. So definitively it is
volume at first place.
But this goes in contradiction to fuel consumption, which is usually
expressed in mass. That is because the engines consume defined mass of fuel
to produce defined power.
This issue is much more complex that may appear at first. In ARINC-429 flow
is transmitted in volume, but also a separate label with fuel density exist.

My guess is that we go with volume as it is much more natural to most ECUs
and FADECs, but we also have a fuel temperature. So calculations into mass
might be done after if the ECU supplies us with both parameters.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/UAVCAN/dsdl/pull/31#issuecomment-311417234>, or mute
the thread

EShamaev:
No fuel injector can use mass. It opens orifice for a specified time and depending on pressure and viscosity certain amount of volume passes.
The ECU can model the mass knowing properties and temperature of fuel.
And mass is actually what ECU should be concerned about to run engine efficiently.

uavcan/equipment/ice/reciprocating/1120.Status.uavcan
```diff
@@ -0,0 +1,178 @@
...
+# Unit: revolutions per minute.
+#
+uint17 engine_speed_rpm
+
+#
+# Estimated lambda coefficient.
+# This parameter is mostly useful for monitoring and tuning purposes.
+# Unit: dimensionless ratio
+#
+float16 lambda_coefficient
```
EShamaev: I feel that lambda should go hand-to-hand with EGT and go into Cylinder Status part

pavel-kirienko: Do they install lambda sensors per cylinder? If that is the case, then sure. Can you please relocate it while you're at it?

uavcan/equipment/ice/reciprocating/1120.Status.uavcan
```diff
@@ -0,0 +1,178 @@
...
+#
+uint22 FLAG_FUEL_PRESSURE_SUPPORTED             = 256
+uint22 FLAG_FUEL_PRESSURE_BELOW_NOMINAL         = 512   # Under-pressure warning
+uint22 FLAG_FUEL_PRESSURE_ABOVE_NOMINAL         = 1024  # Over-pressure warning
+#
+# Detonation and misfire. These flags are optional; either none of them or all of them are supported.
+#
+uint22 FLAG_DETONATION_SUPPORTED                = 2048
+uint22 FLAG_DETONATION_OBSERVED                 = 4096  # Detonation condition observed warning
+uint22 FLAG_MISFIRE_OBSERVED                    = 8192  # Misfire condition observed warning
```
pavel-kirienko: Unlike the rest of the flags here, those pertaining to detonation and misfire represent not a persistent state, but momentary events. Therefore, it should be explicitly specified how long each flag should be maintained in the high state after the event has been observed. My proposition is to require the publisher to clear the flag immediately after broadcasting is done; in this case, a set flag would mean that there was at least one event observed since the previous status message was broadcasted.

EShamaev: Agree.

uavcan/equipment/ice/reciprocating/1120.Status.uavcan
EShamaev: I am not sure how the lambda is installed on planes at all. I have seen it only few times in my life and that was per-cylinder based. Usually there is no such sensor at all on real planes.
So this might be a question to discuss.
I am not sure that moving lambda to cylinders is right, however I am very sure that we can have several exhaust pipes from different cylinders and one sensor can cover only few of the available cylinders.
The question is what we have lambda for. If it is simply for average monitoring - it's fine to leave as it is, but if it is to check for fuel burn conditions and efficiency - should be more than one instance.
I prefer to have ECU manufacturers to comment on this.

EShamaev:
Also I would add chirp (debris) present warning flag for larger engines that could have such sensor installed.
But that's a question to discuss.

pavel-kirienko:
> I am very sure that we can have several exhaust pipes from different cylinders and one sensor can cover only few of the available cylinders.

In that case, the lambda values will be provided only for those cylinders whose exhaust pipes go through the sensor. Makes perfect sense to me. I suggest we move this field, and then wait for criticism from ECU vendors.

> Also I would add chirp (debris) present warning flag for larger engines that could have such sensor installed.

Sure. Flags cost nothing, so we can add plenty of them. In fact, we might need to extend the flag variable to 30 bits wide, for extra future proofness.

@EShamaev will you be able to address this and other concerns we discussed?

uavcan/equipment/ice/reciprocating/1120.Status.uavcan
```diff
@@ -0,0 +1,178 @@
...
+uint22 FLAG_CRANKSHAFT_SENSOR_ERROR_SUPPORTED   = 2
+uint22 FLAG_CRANKSHAFT_SENSOR_ERROR             = 4
+#
+# Temperature levels. These flags are optional; either none of them or all of them are supported.
+#
+uint22 FLAG_TEMPERATURE_SUPPORTED               = 8
+uint22 FLAG_TEMPERATURE_BELOW_NOMINAL           = 16    # Under-temperature warning
+uint22 FLAG_TEMPERATURE_ABOVE_NOMINAL           = 32    # Over-temperature warning
+uint22 FLAG_TEMPERATURE_OVERHEATING             = 64    # Critical overheating
+uint22 FLAG_TEMPERATURE_EGT_ABOVE_NOMINAL       = 128    # Exhaust gas over-temperature warning
```
pavel-kirienko: Misaligned comment (`#`) :scream:

uavcan/equipment/ice/reciprocating/1120.Status.uavcan
```diff
@@ -0,0 +1,178 @@
...
+#
+# Fuel pressure. These flags are optional; either none of them or all of them are supported.
+#
+uint22 FLAG_FUEL_PRESSURE_SUPPORTED             = 256
+uint22 FLAG_FUEL_PRESSURE_BELOW_NOMINAL         = 512   # Under-pressure warning
+uint22 FLAG_FUEL_PRESSURE_ABOVE_NOMINAL         = 1024  # Over-pressure warning
+#
+# Detonation and misfire. These flags are optional; either none of them or all of them are supported.
+#
+uint22 FLAG_DETONATION_SUPPORTED                = 2048
```
pavel-kirienko: We need either `FLAG_DETONATION_SUPPORTED` plus `FLAG_MISFIRE_SUPPORTED`, or `FLAG_DETONATION_MISFIRE_SUPPORTED`. Right now we don't have a flag for `FLAG_MISFIRE_OBSERVED`.

EShamaev: That's tightly coupled with detonation, so I left it in one group.
Will add one more support flag

pavel-kirienko: If they are tightly coupled, then we should probably just rename `FLAG_DETONATION_SUPPORTED` into `FLAG_DETONATION_MISFIRE_SUPPORTED`.

EShamaev: Let's separate

EShamaev:
And we did not get into magneto stuff yet 😄

pavel-kirienko:
I have renamed the chirps warning into debris warning because the original naming seemed like a misnomer. Let me know if this is wrong, I can change it back.

The flags variable was increased to 30 bits; we have 10 bits unused at the moment. I have also added a reserved field 16 bits wide for future use.

yyl123hf:
@frizensami OK,we agreeable with volumetric units.We do not have a fuel density calculation model yet.  Later we will add this part of the algorithm.

uavcan/equipment/ice/reciprocating/1120.Status.uavcan
```diff
@@ -0,0 +1,199 @@
...
+# The index of the publishing ECU.
+#
+uint6 ecu_index
+
+#
+# Spark plug activity report.
+# Can be used during pre-flight tests of the spark subsystem.
+#
+uint3 spark_plug_usage
+#
```
frizensami: @EShamaev i'm not very experienced with engines, could you mind explaining to me what "spark plug usage" is meant to indicate? My best guess is that this is to differentiate between single and dual ignition systems. I suppose it's not for indicating which spark plugs _per cylinder_ are active?

EShamaev: It is indicating which spark plug is active.
image
Usually on double spark plug engines on engine switch you have positions:
OFF-LEFT-RIGHT-BOTH-START
That gives you the possibility to test both spark plugs on ground before takeoff.

frizensami: Ah okay, that's fascinating, thank you.

frizensami:
Okay, all the definitions look good to me :)

pavel-kirienko:
Let us assign a deadline on **Jul 5th**. If there are no further comments and suggestions by that date, we will merge this pull request.

@EShamaev I would like you to confirm that the fuel tank status message makes sense. Thanks in advance.

frizensami:
> Let us assign a deadline on Jul 5th. If there are no further comments and suggestions by that date, we will merge this pull request.

Okay, sounds good, thanks!

EShamaev:
The fuel tank level is fine. Not many things can be there.

pavel-kirienko:
Going to merge this tomorrow. Everyone should have a fresh look.

frizensami:
LGTM :+1:

Issue #3, opened 2015-08-18
Future: Foreign object serialization formats


pavel-kirienko:
The need for this feature is still unclear, but it's worth considering.

An obvious and rather direct approach is to define a message type containing a raw block of bytes with encoded message. The message itself could be encoded in one of the existing self-defining protocol formats, e.g. [Google Protobuf](https://developers.google.com/protocol-buffers/?hl=en), [BSON](https://en.wikipedia.org/wiki/BSON), [etc](https://en.wikipedia.org/wiki/Comparison_of_data_serialization_formats).

Example:

``` python
void1
uint7 RECIPIENT_BROADCAST = 0
uint7 recipient
Name name
Encoding encoding
uint8[<=1024] payload
```

Where `Name`:

``` python
void2
uint8[<64] name
```

`Encoding`:

``` python
uint8 FORMAT_GOOGLE_PROTOBUF = 0
# Other formats...
uint8 format
```

A similar approach can be used for self-defining service calls.

antoinealb:
I think ProtoBuf is not self defining, you have statically define
messages, like in UAVCAN. Another good encoding that we use for our
robots is MessagePack (http://msgpack.org).

LorenzMeier:
@pavel-kirienko @antoinealb You will want to look at this issue:
https://github.com/mavlink/v2_playground/issues/1

It has a complete table. We're currently revisiting the whole message definition on the vehicle in collaboration with Dronecode & 3DR. There was also some interest to have the same messages on CAN, so it would be a big plus to have native support in UAVCAN for the message format. So far Cap'n Proto seems to lead in terms of features and efficiency.

pavel-kirienko:
First stab, at 3DR's request: https://github.com/UAVCAN/dsdl/tree/foreign_object/uavcan/foreign_object

magicrub:
@pavel-kirienko I'm interested in reviving this foreign_object concept for carrying multiple protocols over CAN. I see your branch with the generic 1024 byte buffer. I'm surprised FORMATS does not contain MAVLink? I think that's worthwhile. I definitely have a need for it.

antoinealb:
You can add it to `Format.uavcan`. However I wonder if there is really interest in putting this kind of support in the standard, as the decoding of the payload will probably be application specific anyway.

magicrub:
[...]

PR #39, opened 2018-03-14
Foreign object

magicrub:
This closes issue https://github.com/UAVCAN/dsdl/issues/3.

Add a foreign object packet type that includes a format for:
RAW
BSON
CAPNPROTO
MAVLINK
PROTOBUF

My main goal is to get MAVLink over UAVCAN for a flight controller that only has CAN attached to it (no serial) so a GCS can at least communicate with it without reinventing all of MAVlink cmds over CAN.

PR #40, opened 2018-03-22
High-level protocol tunneling

pavel-kirienko:
Resolves https://github.com/UAVCAN/dsdl/issues/3, supersedes https://github.com/UAVCAN/dsdl/pull/39.

This is a revised implementation of the old concept that takes into account meaningful usage scenarios and the recent improvements of the protocol.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: python
+python:
+  - "3.6"
+before_install:
+  - pip install 'uavcan>=1.0.0.dev27'
+script:
+  - python3 test.py
--- a/README.md
+++ b/README.md
@@ -1,10 +1,13 @@
 Standard DSDL definitions
 =========================
 
-This repository contains DSDL definitions of the standard UAVCAN messages and services.
+[![Build Status](https://travis-ci.org/UAVCAN/dsdl.svg?branch=master)](https://travis-ci.org/UAVCAN/dsdl)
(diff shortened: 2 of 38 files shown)


magicrub:
Looks like you've renamed it to "tunnel" but there are still references to "foreign object" in the source.

pavel-kirienko:
Good catch! Fixed

magicrub:
can you cherry pick this?
https://github.com/magicrub/dsdl/commit/3a0296409ec05f62f6a9bf179e716ec8d307d5a6

We have a ton of enum space, might as well throw a few protocols up there to start with.

pavel-kirienko:
Do you have specific use cases in mind? We can add them, sure (all except `RAW` - that option doesn't make sense and it was a mistake), but that doesn't mean we should. MAVLink is a good option for tunneling, as it is largely self-describing (meaning that you don't need additional context to understand MAVLink traffic); whereas other options are a lot more low-level - they provide mere encoding format rather than a high-level data exchange protocol. Therefore I propose to push back on that; we will always be able to add additional protocols later as necessary.

magicrub:
Nothing specific in mind. The point is to let anyone pass whatever they want. You're arbitrarily trying to restrict that. Doesn't make sense. I don't see any harm filling the enum with the top 50 protocols, whatever they are.

pavel-kirienko:
It is true that I am intentionally trying to restrict the available options because that makes interoperability difficult.

PR #41, opened 2018-04-10
Added message for optic flow sensor data

nikhil9:
A new message type for optic flow sensor.

### DIFF
--- a/uavcan/equipment/optic_flow/1140.OpticFlow.uavcan
+++ b/uavcan/equipment/optic_flow/1140.OpticFlow.uavcan
@@ -1,0 +1,25 @@
+#
+# Optic flow raw data
+#
+
+uavcan.Timestamp timestamp
+
+#
+# raw optic flow data in x
+#
+int16 raw_flow_x
+
+#
+# raw optic flow data in y
+#
+int16 raw_flow_y
+
+#
+# quality
+#
+uint8 qaulity
+
+#
+# angular velocity in sensor frame
+#
+float32[3] optic_flow_gyro


pavel-kirienko:
Could you please describe a use case for this data type that requires the sender to report raw values rather than higher-level data in sensible units (meters)? UAVCAN is trying to be a higher-level protocol, we tend to avoid raw values because they require a lot of context.

I suggest to consider updating this data type to report metric position information. Then again, we already have [uavcan.navigation.GlobalNavigationSolution](https://github.com/UAVCAN/dsdl/blob/master/uavcan/navigation/2000.GlobalNavigationSolution.uavcan).

nikhil9:
Raw values from optic flow sensor are pixel velocities which are further fused(generally EKF) with distance sensor and rotation sensor to get metric velocities. This calculation is mostly done outside sensor hence I am proposing raw. For convenience I can drop raw from variable and rename it to pixel_flow if that helps. Let me know. Thanks.

pavel-kirienko:
> For convenience I can drop raw from variable and rename it to pixel_flow if that helps

It does, please do.

You should also specify the units of measurement for the x and y fields.

There's a typo in `quality`; also its units of measurement are missing. I suggest using a regular error variance in pixel², does that make sense? Perhaps a covariance matrix is needed for generality?

Do we not need a covariance matrix for the angular velocity readings? Also I suggest to make the naming more explicit and name it `sensor_angular_velocity`.

@mhkabir perhaps you could provide some feedback as well?

mhkabir:
I would suggest creating a more advanced version of the Mavlink optical flow message. It is mature and has proven integrations with standard navigation estimators.
http://mavlink.org/messages/common#OPTICAL_FLOW_RAD

Suggested skeleton message : 
```
timestamp
integration_time
integrated_flow_x (rad)
integrated_flow_y (rad)
integrated_flow_z (rad)
flow_covariance
integrated_xgyro (rad)
integrated_ygyro (rad)
integrated_zgyro (rad)
integrated_xaccel (m/s)
integrated_yaccel (m/s)
integrated_zaccel (m/s)
```

pavel-kirienko:
```
integrated_flow_x (rad)
integrated_flow_y (rad)
integrated_flow_z (rad)
```

What does `rad` represent here?

mhkabir:
Radians.

pavel-kirienko:
Ah I see, you represent both angular and linear displacement using the same quantity since they are observed equivalently by the sensor. Although this method seems to require a distance estimate between the sensor and the flow field, which is missing from your definition. Or am I misunderstanding something?

Also, how do you define flow around the Z axis?

mhkabir:
> Ah I see, you represent both angular and linear displacement using the same quantity since they are observed equivalently by the sensor.

Yep. This does require a distance estimate to the flow field, but we normally use a rangefinder to estimate that, and a timestamped range message should be used for it.

> how do you define flow around the Z axis?

Yes, this needs some thought. I'm not sure yet how to define the "flow", but I was thinking about a yaw rotation estimate from a flow sensor, which makes sense. Any ideas?

pavel-kirienko:
> This does require a distance estimate to the flow field, but we normally use a rangefinder to estimate that, and a timestamped range message should be used for it.

So, a separate message?

> I was thinking about a yaw rotation estimate from a flow sensor, which makes sense.

It doesn't fit the same model used for X/Y estimates. Linear displacement along the Z axis is manifested as a change of scale on the flow field; whereas angular displacement is seen as a rotation of the field. Both can be observed independently, so we're going to need two separate states for linear and angular Z readings. Unless I'm missing something.

mhkabir:
> So, a separate message?

Yes. Probably https://github.com/UAVCAN/dsdl/blob/master/uavcan/equipment/range_sensor/1050.Measurement.uavcan?

> It doesn't fit the same model used for X/Y estimates. Linear displacement along the Z axis is manifested as a change of scale on the flow field; whereas angular displacement is seen as a rotation of the field. Both can be observed independently, so we're going to need two separate states for linear and angular Z readings. Unless I'm missing something.

Yes this makes sense. However I'm not sure how useful the linear Z reading will be since it is unscaled (not metric). The rotational component however, is very useful in practice since it constrains yaw estimation.

pavel-kirienko:
I suggest this then:

```python
uavcan.Timestamp timestamp

uint8 sensor_id

float32 integration_time        # [second]

#
# Sensor RH rotation about the X axis induces a positive flow on X.
# Sensor linear motion along the positive Y axis induces a negative flow on X.
#
# Sensor RH rotation about the Y axis induces a positive flow on Y.
# Sensor linear motion along the positive X axis induces a positive flow on Y.
#
float32[2] integrated_flow_xy   # Flow in radians around X/Y [rad]
float32 integrated_flow_z       # Rotation of the flow field [rad]

float32[3] integrated_angvel    # X, Y, Z [rad]

float32[3] integrated_accel     # X, Y, Z [m/s]

float16[<=9] flow_covariance    # X, Y, Z [rad^2 along the diagonal, rad elsewhere]
```

@nikhil9 could you please update your PR with this if there are no objections?

pavel-kirienko:
Closing as stale.

PR #44, opened 2018-06-16
Add explicit version to all messages

kjetilkjeka:
I've verified that the rust dsdl parser and compiler have basic support for version numbers. I've not tested with multiple versions of same filename but think this will fail. 

We should at least hold on merging this PR until the python parser have the same basic support for version numbers.

---

 - [x] Verify that the rust dsdl parser is able to parse dsdl when there only exists a single version of a message. 
 - [x] Verify that the rust dsdl compiler is able to compile dsdl when there only exists a single version of a message.
 - [ ] Implement parsing of version number in python parser. Uavcan/pyuavcan#39
 - [ ] Ensure that dsdl compilation is done correctly for pyuavcan when there only exists a single version of a message.
 - [ ] Ensure that dsdl compilation is done correctly for libuavcan when there only exists a single version of a message.

Possibly:

 - [ ] Figure out how compilation should be done when there exists several versions of same message.
 - [ ] Ensure that dsdl compilation is done correctly for pyuavcan when there exists a several versions of the same message.
 - [ ] Ensure that dsdl compilation is done correctly for libuavcan when there exists a several versions of the same message.
 - [ ] Ensure that dsdl compilation is done correctly for uavcan.rs when there exists a several versions of the same message.


kjetilkjeka:
My first thought on how to compile multiple version for the Rust compiler is:
 - Append `Ver<major version>` to the typename for every major version.
 - Append `Ver0<minor version>` to the typename for every minor version that has 0 as its major version.
 - Add a type alias from the type name to the most recent version.

I do not know the details of how libuavcan compile dsdl but I guess this is possible here as well? Do you have any better ideas  @pavel-kirienko ?

kjetilkjeka:
Perhaps we instead should make `NodeStatus` version 1.0? Since this message is required to support the Uavcan protocol it would perhaps be valuable that only one version exists? This would of course mean that we're commited to not change this message, which i assume is not a problem?

pavel-kirienko:
I don't think we can arrive at any sensible conclusion here until we decided on how to detect data structure version at the time of decoding; and, optionally, whether to attempt automatic promotion of received data structures to the latest supported version in libuavcan, uavcan.rs, and pyuavcan. Let's finalize https://github.com/UAVCAN/dsdl/issues/35 first.

kjetilkjeka:
Ok, I'm on board with that. I've been actually doing some work on this front today. And i wrote a half baked solution there just now. I will refine it a bit more a write a throrough proposal very soon.

pavel-kirienko:
@kjetilkjeka I started working on a new set of DSDL definitions for UAVCAN v1.0. I suggest closing this in favor of my new proposition which will incorporate this change as well.

PR #48, opened 2018-10-23
Diagnostic namespace

pavel-kirienko:
This is my new controversial diagnostic message. I have modified it compared to the original published in the previous PR to accommodate some of the feedback from @kjetilkjeka. I decided to keep my original set of severity levels more or less unchanged, pending your review. I added a bit of extra information in the comments to somewhat justify my decisions.

Questions specifically for @thirtytwobits: 

* Does the phrase "severity level" make sense outside of the CS context?
* Does the phrase "diagnostic record" make sense in the context of human-readable messages for logging (i.e., non real-time processing) and displaying (in real-time) purposes?

### DIFF
--- a/uavcan/diagnostic/65520.Record.1.0.uavcan
+++ b/uavcan/diagnostic/65520.Record.1.0.uavcan
@@ -1,0 +1,17 @@
+#
+# Generic human-readable text message for logging and displaying purposes.
+# Generally it should be published at the lowest priority level.
+#
+
+# Optional timestamp; zero if undefined.
+uavcan.time.Point.1 timestamp
+
+# Standard severity, 3 bit wide.
+Severity.1 severity
+
+# Short name of the message source within the publishing node.
+uint8[<32] source
+
+# The actual useful text of the message.
+# Normally, messages should be kept as short as possible, especially those of high severity.
+uint8[<256] text
--- a/uavcan/diagnostic/65520.Record.1.0.uavcan
+++ b/uavcan/diagnostic/65520.Record.1.0.uavcan
@@ -3,7 +3,8 @@
 # Generally it should be published at the lowest priority level.
 #
 
-# Optional timestamp; zero if undefined.
+# Optional timestamp in the network-synchronized time system; zero if undefined.
+# The timestamp value conveys the exact moment when the reported event took place.
 uavcan.time.Point.1 timestamp
 
 # Standard severity, 3 bit wide.
--- a/uavcan/diagnostic/65520.Record.1.0.uavcan
+++ b/uavcan/diagnostic/65520.Record.1.0.uavcan
@@ -12,6 +12,10 @@ Severity.1 severity
 # Short name of the message source within the publishing node.
 uint8[<32] source
 
+@assert offset % 8 == {0}
+
 # The actual useful text of the message.
 # Normally, messages should be kept as short as possible, especially those of high severity.
 uint8[<256] text
+
+@assert offset % 8 == {0}
--- a/uavcan/diagnostic/65520.Record.1.0.uavcan
+++ b/uavcan/diagnostic/65520.Record.1.0.uavcan
@@ -5,7 +5,7 @@
 
 # Optional timestamp in the network-synchronized time system; zero if undefined.
 # The timestamp value conveys the exact moment when the reported event took place.
-uavcan.time.Point.1 timestamp
+uavcan.time.SynchronizedTimestamp.1 timestamp
 
 # Standard severity, 3 bit wide.
 Severity.1 severity
--- a/uavcan/diagnostic/Severity.1.0.uavcan
+++ b/uavcan/diagnostic/Severity.1.0.uavcan
@@ -1,0 +1,39 @@
+#
+# Generic message severity representation.
+#
+
+# The severity level ranging from 0 to 7, where low values represent low-severity (unimportant) messages, and
+# high values represent high-severity (important) messages. Several mnemonics for the severity levels are
+# defined below. They differ quite a bit compared to the traditional mnemonics widely used in conventional
+# software systems in order to ensure conceptual compatibility with the rest of the UAVCAN specification (see
+# the standard message type uavcan.node.Heartbeat), and to emphasize the much higher cost of the medium
+# compared to regular local log files. Severity levels that don't have an associated mnemonic name should not
+# be used.
+uint3 severity
+
+# Unimportant, low-impact messages related to expected periodic events and processes.
+# Normally, messages of this level should never be enabled in deployed/in-field applications.
+uint3 TRIVIAL = 0
+
+# Messages related to expected events and unimportant decisions that can be safely ignored during normal operation.
+# Generally, messages of this severity level should not be published more often than once per second on average.
+# Normally, messages of this level (and lower) should be suppressed unless specifically requested otherwise.
+uint3 LOW = 1
+
+# Severity level 2 is reserved for future use.
+
+# Noteworthy messages about important or rare events and decisions
+# Generally, messages of this severity level should not be published more often than once per 10 seconds on average.
+# Normally, messages of this level (and lower) should be suppressed unless specifically requested otherwise.
+uint3 MODERATE = 3
+
+# Severity level 4 is reserved for future use.
+# Severity level 5 is reserved for future use.
+
+# High-impact information about unexpected events, unusual conditions, malfunctions, or errors.
+# Messages of this priority level (and higher) should be enabled (not suppressed) by default.
+uint3 HIGH = 6
+
+# Exceptionally rare circumstances or severe failures.
+# Disabling (suppressing) messages at this severity level is not recommended.
+uint3 EXCEPTIONAL = 7
--- a/uavcan/diagnostic/Severity.1.0.uavcan
+++ b/uavcan/diagnostic/Severity.1.0.uavcan
@@ -28,11 +28,12 @@ uint3 LOW = 1
 uint3 MODERATE = 3
 
 # Severity level 4 is reserved for future use.
-# Severity level 5 is reserved for future use.
 
 # High-impact information about unexpected events, unusual conditions, malfunctions, or errors.
 # Messages of this priority level (and higher) should be enabled (not suppressed) by default.
-uint3 HIGH = 6
+uint3 HIGH = 5
+
+# Severity level 6 is reserved for future use.
 
 # Exceptionally rare circumstances or severe failures.
 # Disabling (suppressing) messages at this severity level is not recommended.
--- a/uavcan/diagnostic/Severity.1.0.uavcan
+++ b/uavcan/diagnostic/Severity.1.0.uavcan
@@ -4,37 +4,39 @@
 
 # The severity level ranging from 0 to 7, where low values represent low-severity (unimportant) messages, and
 # high values represent high-severity (important) messages. Several mnemonics for the severity levels are
-# defined below. They differ quite a bit compared to the traditional mnemonics widely used in conventional
-# software systems in order to ensure conceptual compatibility with the rest of the UAVCAN specification (see
-# the standard message type uavcan.node.Heartbeat), and to emphasize the much higher cost of the medium
-# compared to regular local log files. Severity levels that don't have an associated mnemonic name should not
-# be used.
-uint3 severity
+# defined below. Nodes are advised to implement output filtering mechanisms, allowing users to select
+# the minimal severity for emitted messages; messages of the selected and higher severity levels will
+# be published, and messages of lower severity will be suppressed (discarded).
+uint3 value
 
-# Unimportant, low-impact messages related to expected periodic events and processes.
-# Normally, messages of this level should never be enabled in deployed/in-field applications.
-uint3 TRIVIAL = 0
+# Messages of this severity can be used only during development.
+# They must not be used in a fielded operational system.
+uint3 TRACE = 0
 
-# Messages related to expected events and unimportant decisions that can be safely ignored during normal operation.
-# Generally, messages of this severity level should not be published more often than once per second on average.
-# Normally, messages of this level (and lower) should be suppressed unless specifically requested otherwise.
-uint3 LOW = 1
+# Messages that can aid in troubleshooting.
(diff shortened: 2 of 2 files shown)


uavcan/diagnostic/65520.Record.1.0.uavcan
```diff
@@ -0,0 +1,17 @@
+#
+# Generic human-readable text message for logging and displaying purposes.
+# Generally it should be published at the lowest priority level.
+#
+
+# Optional timestamp; zero if undefined.
+uavcan.time.Point.1 timestamp
```
kjetilkjeka: The changes in #50 makes this more sensible as:
 - We have an explicit timeframe meaning we can plot the time and date in a human friendly way
 - We can send a "Time since boot" `SystemTime` as a fallback when no time sync exists, allowing us to remove the notion of a special for zero.

kjetilkjeka:
The CAN bus is a limited resource and I assume the severity exists to be able to filter on it. We should have a standard service for changing the severity threshold of what messages that are actually being sent over the bus.

I have some questions.
1. Can you describe this other use case than logging you where talking about? Realtime diagnostics I think you called it.
2. What is the reason we don't want to have seperate types and default subjects for logging and realtime diagnostics. Isn't one of the ideas with the new system that having several subjects for similar stuff is cheap and efficient.
3. How do a dev that wants a standard logging interface use the severity level? I have the following use cases:
     - I want to have trace log messages (that can be enabled/disabled) for use in bench testing, at the same time I also want debug level (for enabling in production when weird bugs occur) and info messages that can be enabled depending on different factors. I want to be able to dynamically chose a log level that only sends the messages i need on the bus. 
     - I'm using the uavcan diagnostics stuff for the first time. How do I find out how to map the semantics of the log messages the rest of the world uses  (`info`, `warn`, `error`, etc) into these types.

thirtytwobits:
- Does the phrase "severity level" make sense outside of the CS context?

It's a bit redundant. `severity` is adequate.

- Does the phrase "diagnostic record" make sense in the context of human-readable messages for logging (i.e., non real-time processing) and displaying (in real-time) purposes?

Yep, although "record" implies (but does not specify) something that is stored and can be retrieved. "log" or "message" implies ephemeral data.

pavel-kirienko:
@kjetilkjeka 

> Can you describe this other use case than logging you where talking about? Realtime diagnostics I think you called it.

Displaying important messages on the screen for the human operator/technician/developer.

> What is the reason we don't want to have seperate types and default subjects for logging and realtime diagnostics. Isn't one of the ideas with the new system that having several subjects for similar stuff is cheap and efficient.

Because they are too similar and there is no reason to force developers to choose between them. One would expect that something that is important enough to be displayed is also worth saving in the log.

> How do a dev that wants a standard logging interface use the severity level? I have the following use cases:

For now I would suggest setting the output level via a configuration parameter. Eventually we could either define a standard service for setting the output level, or (preferred) define a standard register (parameter) name for that, e.g. `uavcan.diagnostic.level`.

Regarding the semantics: I put some guidelines in the comments of the severity level message. If they are still not helpful enough, we could debate either extending them, or going back to the more traditional trace/debug/info/notice/warning/error/critical scheme. What I seriously dislike there is that we'd have to throw out the conventions adopted for the Heartbeat message. Perhaps we should have a vote?

thirtytwobits:
> Displaying important messages on the screen for the human operator/technician/developer.

If this is a use-case then we should use FAA required terminology for this (i.e. the new terminology used in NodeStatus)

So, does this replace the logging message we had in the debug namespace?

pavel-kirienko:
> So, does this replace the logging message we had in the debug namespace?

Yes.

kjetilkjeka:
> Displaying important messages on the screen for the human operator/technician/developer.

I still don't like that this obfuscates a feature which more or less everyone is going to want to use. Easy to use logging is realstically one of the most productivity enhancing features in uavcan.

> Because they are too similar and there is no reason to force developers to choose between them. One would expect that something that is important enough to be displayed is also worth saving in the log.

This relationship only holds one way though, you might not want every item in the log on your RT diagnostics screen. The FAA terms also seems to be worried about how things are indicated to the operator, and not much on logging. I have a solution in mind.

Since logging is a thing you always want to do, you always have to provide a sensible log level. If you also want to display the message on a screen or indicator, you provide a severity as well.

```python
#
# Generic human-readable text message for logging and displaying purposes.
# Generally it should be published at the lowest priority level.
#

#
# A record should only be sent if it matches the nodes configured log level or indication severity.
#

# The time reference in use
TimeReference.0.1 reference

# The duration since the provided time reference (use boot if no time sync exists)
truncated uint53 timestamp

# Log level (Trace, Debug, Info, Warn, Error, Fatal) TODO: create const defs
#
# Is used to decide if message should be logged, and may be used when filtering logs.
# Can be thought of as an indication to the developer/system administrator.
#
# If the message is only relevant for the operator, `Info`, is a pretty safe log level to use.
uint3 log_level

# Optional standard indication_severity (None, Trivial, Low, Moderate, High, Exceptional) TODO: create const defs
#
# Is used to decide if/how a record is indicated to the operator.
# Use `None` if a record should only be logged and never displayed.
uint3 indication_severity

void7

# Short name of the message source within the publishing node.
uint8[<32] source
@assert offset % 8 == {0}

# The actual useful text of the message.
# Normally, messages should be kept as short as possible, especially those of high severity.
uint8[<256] text
@assert offset % 8 == {0}
```

One advtage of my proposal is that it's pretty clear how things like trace/debug messages should be defined (Trace/debug + None) and still pretty clear how things like operator indication messages should be sent (Info + severity)

pavel-kirienko:
```
# The time reference in use
TimeReference.0.1 reference

# The duration since the provided time reference (use boot if no time sync exists)
truncated uint53 timestamp
```

Is this a good case study for the time representation problem or what? There may be no more than two valid options:

- A proper timestamp that belongs to the network-synchronized time system.
- A node-local monotonic timestamp as a duration since boot.

Adding the full TimeReference selector seems counterproductive because not only it eats bits, it adds invalid states: suppose the network-synchronized time is ApplicationSpecific, and the reported timestamp contains TAI time. Receiving nodes would have to add mandatory logic to take care of possible time discrepancies, and, perhaps, conversions between well-defined time systems (e.g. UTC and TAI, provided that the number of leap seconds is known). If the ability to report optional node-local time is much desired, there should be just a single bit selecting between node-local and network-synchronized time. But even that is bad because of the extra variance that doesn't seem to add much value: a timestamp that belongs to a remote node's local time system makes a very limited sense anywhere outside that node, unless highly complex methods of clock recovery are involved (e.g. the Olson's algorithm). I suggest that in this and similar circumstances we just stick to a simple naive timestamp value that is either a valid network-synchronized time or nothing. The "nothing" option could be represented either as a zero or a dedicated selector bit; I am inclined to prefer the former.

-----

The two levels of importance (log level and severity) side by side do look like an over-engineering. I understand that the motivation of this proposal is to bring my original design closer to a form that is more conventional and more accepted by other developers. If we failed to agree on using my suggested set of severity levels, we should just discard it and go back to the standard set of TRACE/DEBUG/INFO/etc. rather than adding an orthogonal dimension of importance level next to it.

kjetilkjeka:
> Adding the full TimeReference selector seems counterproductive because not only it eats bits, it adds invalid states: suppose the network-synchronized time is ApplicationSpecific, and the reported timestamp contains TAI time. Receiving nodes would have to add mandatory logic to take care of possible time discrepancies, and, perhaps, conversions between well-defined time systems (e.g. UTC and TAI, provided that the number of leap seconds is known). 

I wouldn't expect this message to be used for anything related to control, only logging (display/writing to file/read by human sorts of things). This means that any log software could eat these messages without any context (what you call network-synchronized time) and plot them correctly in human readable time.

If application-specific time is used it wouldn't be able to plot it directly in human readable time, but it would have explicit information that the time reference is unknown and might request it. Even if it is not available it could correctly sort the log and display time since unknown reference.

I think all this is great, and would only lead to an error if the logging application was trying to also read time sync data and expected them to be the same time reference. I think this is a dangerous assumption remaining from before there was a possibility to explicitly define a time frame.  The error of sending a timestamp in a different time frame than what's expected is actually only possible because of this assumption that they will be the same. And when explicitly defining the time frame for log messages this kind of mismatch errors are much more unlikely to happen.

On the other hand, some things used for control should probably be given in network synchronized time. I don't think being explicit about the time reference in the case of logging contradicts this. I only think it supports my claim that timestamping is different enough that it will be hard to create types that fit in all cases.

> If the ability to report optional node-local time is much desired, there should be just a single bit selecting between node-local and network-synchronized time. But even that is bad because of the extra variance that doesn't seem to add much value: a timestamp that belongs to a remote node's local time system makes a very limited sense anywhere outside that node, unless highly complex methods of clock recovery are involved (e.g. the Olson's algorithm). I suggest that in this and similar circumstances we just stick to a simple naive timestamp value that is either a valid network-synchronized time or nothing. The "nothing" option could be represented either as a zero or a dedicated selector bit; I am inclined to prefer the former.

It's better to have a way of knowing how much time between errors occurred than not knowing at all. With the explicit notion of this being time since boot, it's unlikely that any damage will be done from mistaking it for a global time reference.

It's not necessary for the time since boot timestamp to make sense, and I'm not sure how Olson's algorithm work, but wouldn't you only need to add a service that asked the node the time of boot in a different time frame to be able to backtrack the log messages received before the first time sync?

pavel-kirienko:
Sorry, I failed to specify the domain to which my previous post should apply to. I was talking about the problem of message timestamping in general, not just this specific case of log records.

I suggest we move this time-related discussion to https://github.com/UAVCAN/dsdl/pull/50.

pavel-kirienko:
> If we failed to agree on using my suggested set of severity levels, we should just discard it and go back to the standard set of TRACE/DEBUG/INFO/etc. rather than adding an orthogonal dimension of importance level next to it.

I want to be redundant and ask @UAVCAN/developers again if we want this new system of severity levels proposed in this PR (TRIVIAL, LOW, MODERATE, HIGH, EXCEPTIONAL), or we want conformance with the practices accepted in the software development field at large (DEBUG, INFO, NOTICE, WARNING, ERROR, ERROR, CRITICAL, ALERT, EMERGENCY).

If there are arguments in favor or against of either, please post here. If this is considered to be a matter of taste, please vote on this comment up :+1:  in favor of TRIVIAL..., down :-1: in favor of DEBUG...

pavel-kirienko:
Missing vote from @thirtytwobits

thirtytwobits:
Ugh, I don't like either. So, to be clear, we have both log-level and severity level with the latter being a marker of the severity as displayed to an operator and the former being the "severity" as displayed to a developer? If this is correct then we'd have:

```
# Log level (Trace, Debug, Info, Warn, Error, Fatal)
#
# Is used to decide if message should be logged, and may be used when filtering logs.
# Can be thought of as an indication to the developer/system administrator.
#
# If the message is only relevant for the operator, `Info`, is a pretty safe log level to use.
uint3 log_level

# Optional standard indication_severity (None, Nominal, Advisory, Caution, Warning)
#
# Is used to decide if/how a record is indicated to the operator.
# Use `None` if a record should only be logged and never displayed.
uint3 indication_severity
```

It's a compromise but like all compromises it's flawed. For example, this means we could have messages that were both Fatal and Nominal or Debug and Warning. That's just insane and perhaps dangerous.

We either need to adopt FAA controlled severities or we need two APIs, one for debug logging and one for operator notification. I would actually prefer this option. If so then the debug logging should using Trace, Debug, Info, etc. and the operator notification API should use FAA controlled terms.

pavel-kirienko:
No, the current proposition is to have just one severity dimension for both logging and indication. So the chosen approach (TRIVIAL... or DEBUG...) applies to both.

> We either need to adopt FAA controlled severities or we need two APIs, one for debug logging and one for operator notification.

Is it an option to implement simple translation at the time of displaying? Like:

- DEBUG, INFO, NOTICE --> ADVISORY
- WARNING, ERROR --> CAUTION
- CRITICAL, ALERT, EMERGENCY --> WARNING

thirtytwobits:
> Is it an option to implement simple translation at the time of displaying? Like:
> 
> * DEBUG, INFO, NOTICE --> ADVISORY
> * WARNING, ERROR --> CAUTION
> * CRITICAL, ALERT, EMERGENCY --> WARNING

Sure. We can provide this as a bulletin. I actually like this since we may have different bulletins for EASA, CAA, FAA, etc. One refinement though, to be thorough:

```
 * DEBUG, INFO --> NOMINAL (do not alert the operator)
 * NOTICE --> ADVISORY
 * WARNING, ERROR --> CAUTION
 * CRITICAL, ALERT --> WARNING
```

That all said, using terms like "emergency" is problematic. Determination that something is an "emergency", a "catastrophe", or a "cluster-f**k" (etc) is something that should be done at the system level and should not be something a single component can decide. 

Finally, it's still problematic to me to say that this is an API that might be used to notify operators of events. This might be something we do not want to take on in our API leaving this to a higher level to determine.

So, I guess I'll vote for developer terminology and then we should issue bulletins to provide translation and operator notification guidance.

pavel-kirienko:
Okay, agreed. @thirtytwobits @kjetilkjeka @antoinealb Please review what's below and confirm that we have a consensus:

- TRACE
- DEBUG
- INFO
- NOTICE
- WARNING
- ERROR
- CRITICAL
- ALERT

pavel-kirienko:
I believe this is ready for review.

pavel-kirienko:
I am going to merge this around tomorrow unless someone objects.

kjetilkjeka:
Should we make the timestamping self-contained for this message by including the time reference? It doesn't require any synchronization besides displaying messages in correct order.

On the contrary to messages intended for synchronization it makes also sense to "change" time reference from boot time (before time synchronization) to some known time reference in the middle of operation.

pavel-kirienko:
Selectable time references extend the state set, complicate failure analysis and testing. I just talked about here in this thread earlier. Also, runtime transitions between different time systems add transients that may be very hard to deal with and test. I talked about this in the timestamping PR. Time systems should be static and unchanging. If we want boot time, we can always add a separate optional timestamp instead, but I'm not sure it's the right thing to do in this particular case because it bloats the message and I'm not sure whether it's worth it.

PR #47, opened 2018-10-11
Initial draft proposal of the new standard data type set for UAVCAN v1.0

pavel-kirienko:
This is my vision of the new data type set design built in accordance with [our new approaches](https://forum.uavcan.org/t/stockholm-summit-recap/170). One will notice that this new set is completely devoid of any equipment-specific data types, which is fully intentional. The old approach that we experimented with in v0 turned out to be deeply flawed because it forces one to design their applications in a very particular way, rendering the protocol unusable for use cases that were not foreseen when the set of data types was originally designed. More background info on the subject is available in the forum thread linked above.

The new approach is to provide a set of very generic data types that do not make any assumptions about the nature of the target application. The resulting type system is far less efficient, but the upside is that it is much more flexible and repurposable. Rather than trying to cater to every sensible application out of the box, we are now encouraging equipment vendors to provide custom vendor-specific message definitions for their equipment when higher efficiency or advanced functionality is desired; whereas simpler applications can be handled by using the standard set alone. I expect that there will be equipment that can be interfaced with using both generic data types (in a universal but not-so-efficient way) and vendor-provided data types (in a very vendor-specific but very efficient way). Similar patterns can be found in other systems, e.g. CANopen or USB.

I think it might be sensible to encourage vendors to publish their vendor-specific data types under very permissive licenses permitting other vendors to re-use their data type definitions to reduce fragmentation of the ecosystem. We could also come up with a set of simple domain-specific definitions to be used as a middle ground between the generic standard set and vendor-specific custom messages, although that should belong in a completely separate namespace outside of `uavcan.*`.

I want everyone involved to evaluate the new data type set carefully and pay particular attention to the following parts:

`uavcan.node.GetPorts`, `uavcan.node.GetPortInfo`, `uavcan.node.GetPortStatistics` - these data types provide a simple yet very powerful API that allows one to reconstruct the computational graph, collect the main communication metrics (input/output/errors) in real time, and verify that all nodes use compatible data types for communication with each other. Eventually, when the new GUI tool is around, it will be able to build things like this, showing the traffic intensity and transport errors in real time:

cv_graph

(regarding the term "port", see https://forum.uavcan.org/t/a-generic-term-for-either-subject-or-service/182)

[...]

PR #49, opened 2018-10-23
Remote file system manipulation services

pavel-kirienko:
Some of the context is available in the old PR here: https://github.com/UAVCAN/dsdl/pull/47#issuecomment-431584537

Our objectives are:

* Minimal shared state. The protocol should be client-driven as much as possible, and servers must not be required to maintain context across multiple requests. Although they are not prohibited from doing so either, e.g., for optimization purposes.
* Full CRUD is nice to have.

For reference, to the best of my knowledge, there have been only two uses for this protocol so far:
  * Firmware update (which is fundamentally a read-only operation)
  * Remote log downloading (which is quite similar to firmware update)

Paging @thirtytwobits @davids5 @kjetilkjeka

### DIFF
--- a/uavcan/file/405.GetInfo.0.1.uavcan
+++ b/uavcan/file/405.GetInfo.0.1.uavcan
@@ -24,3 +24,5 @@ void4
 
 # Reserved for future use
 void64
+
+@assert offset % 8 == {0}
--- a/uavcan/file/405.GetInfo.0.1.uavcan
+++ b/uavcan/file/405.GetInfo.0.1.uavcan
@@ -1,0 +1,26 @@
+#
+# Information about a remote file system entry (file, directory, etc).
+#
+
+Path.1 path
+
+---
+
+# Result of the operation
+Error.1 error
+
+# File size in bytes. Should be set to zero for directories.
+truncated uint40 size
+
+# The UNIX Epoch time when the entry was last modified. Zero if unknown.
+truncated uint40 unix_timestamp_of_last_modification
+
+# If such entry does not exist, all flags should be cleared/ignored.
+bool is_file_not_directory  # True if file, false if directory
+bool is_link                # This is a link to another entry; the above flag indicates the type of the target
+bool is_readable            # The item can be read by the caller (applies to files and directories)
+bool is_writeable           # The item can be written by the caller (applies to files and directories)
+void4
+
+# Reserved for future use
+void64
--- a/uavcan/file/407.Move.1.0.uavcan
+++ b/uavcan/file/407.Move.1.0.uavcan
@@ -1,0 +1,29 @@
+#
+# Manipulate a remote file system entry. Applies to files, directories, and links alike.
+# If the remote entry is a directory, all nested entries will be affected, too.
+#
+# Move/Copy
+#   Specify the source path and the destination path.
+#   If the source does not exist, the operation will fail.
+#   Set the preserve_source flag to copy rather than move.
+#   If the destination exists and overwrite_destination is not set, the operation will fail.
+#
+# Touch
+#   Specify the same path in the source and in the destination.
+#   Flags are ignored.
+#
+# Remove
+#   Specify the source path and make the destination path empty.
+#   Flags are ignored.
+#
+
+bool preserve_source            # Do not remove the source. Used to copy instead of moving.
+bool overwrite_destination      # If the destination exists, remove it beforehand.
+void30
+
+Path.1 source
+Path.1 destination
+
+---
+
+Error.1 error
--- a/uavcan/file/409.Write.1.0.uavcan
+++ b/uavcan/file/409.Write.1.0.uavcan
@@ -21,6 +21,9 @@ Path.1 path
 
 uint8[<=128] data
 
+@assert offset % 8 == {0}
+@assert offset.max / 8 <= 300
+
 ---
 
 Error.1 error
--- a/uavcan/file/409.Write.1.0.uavcan
+++ b/uavcan/file/409.Write.1.0.uavcan
@@ -10,10 +10,6 @@
 #
 # When the write operation is complete, the server shall truncate the resulting file past the specified offset.
 #
-# Server implementation advice:
-# It is recommended to implement proper handling of concurrent writes to the same file from different clients, for
-# example by maintaining a staging area for uncompleted writes (like FTP servers do).
-#
 
 truncated uint40 offset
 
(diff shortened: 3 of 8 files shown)


uavcan/file/405.GetInfo.0.1.uavcan
```diff
@@ -0,0 +1,26 @@
...
+---
+
+# Result of the operation
+Error.1 error
+
+# File size in bytes. Should be set to zero for directories.
+truncated uint40 size
+
+# The UNIX Epoch time when the entry was last modified. Zero if unknown.
+truncated uint40 unix_timestamp_of_last_modification
```
kjetilkjeka: #50 `SystemTime`?

uavcan/file/409.Write.1.0.uavcan
```diff
@@ -0,0 +1,26 @@
...
+# Server implementation advice:
+# It is recommended to implement proper handling of concurrent writes to the same file from different clients, for
+# example by maintaining a staging area for uncompleted writes (like FTP servers do).
+#
+
+truncated uint40 offset
+
+Path.1 path
+
+uint8[<=128] data
```
kjetilkjeka: Àny reason this is asymmetrical to `Read`?

uavcan/file/407.Move.1.0.uavcan
```diff
@@ -0,0 +1,29 @@
...
+#   Specify the source path and the destination path.
+#   If the source does not exist, the operation will fail.
+#   Set the preserve_source flag to copy rather than move.
+#   If the destination exists and overwrite_destination is not set, the operation will fail.
+#
+# Touch
+#   Specify the same path in the source and in the destination.
+#   Flags are ignored.
+#
+# Remove
```
kjetilkjeka: The type should probably be called "Modify" instead of Move. As it can both Move, Copy, Touch and Remove.

uavcan/file/407.Move.1.0.uavcan
```diff
@@ -0,0 +1,29 @@
...
+# If the remote entry is a directory, all nested entries will be affected, too.
+#
+# Move/Copy
+#   Specify the source path and the destination path.
+#   If the source does not exist, the operation will fail.
+#   Set the preserve_source flag to copy rather than move.
+#   If the destination exists and overwrite_destination is not set, the operation will fail.
+#
+# Touch
+#   Specify the same path in the source and in the destination.
```
thirtytwobits: Shouldn't touch be that you don't specify a source? (i.e. if the file doesn't exist then create a zero-length file).

pavel-kirienko: Fixed

uavcan/file/407.Move.1.0.uavcan
pavel-kirienko: Fixed

uavcan/file/409.Write.1.0.uavcan
pavel-kirienko: Yes, because we have to transfer both the name and the payload. Having a 256-byte buffer here would result in a structure over 416 bytes large, so I had to pick the next lower power of two. Power of two is important because it is convenient for applications involved with low-level memory access, such as bootloaders.

pavel-kirienko:
> Can we either add a hash to GetInfo or (preferably) send timestamp of last modify action for every read?

We used to have a CRC64 in GetInfo, a very long time ago (in 2015). Got scrapped because it was a great way to hang up the server by requesting GetInfo on a large file.

Conventional file transfer protocols don't seem to bother about this. Adding the timestamp would guard you against some of the race conditions, but may fail to save you from more exotic corner cases like the system time leaping backward. I think if we were to become concerned about integrity, adding a hash is probably the most sensible way to do it.

Now, suppose we want a hash. There are three obvious possibilities that come to mind:

- Add a service for requesting the hash. The callers may have to tolerate extremely long response times when large files are involved.
- Extend GetInfo with an optional hash that has to be requested specifically by the caller. Large files are still a problem.
- Do it the FTP way and follow a convention where the hash (or, even better, a digital signature) is located right next to the file under the same name suffixed with a specific extension (`.asc`, `.md5`, `.pgp`). Example: https://ftp.mozilla.org/pub/firefox/releases/63.0/linux-x86_64/ru/

The latter option seems to be the simplest (as it is a convention, it doesn't require rigorous specification and adherence from implementers) and the most flexible option. For large files that rarely change, the signature can be pre-computed and stored in a file. Some security-minded server implementations may opt to generate hash/signature files on the fly in a manner that is transparent to the client (e.g., check if the file exists; if it does, return it, otherwise create it in-memory and then return).

antoinealb:
We could add a `version` field that would monotonically increase on each file update, so that you could reject chunks that arrive with a different version? This is quite simple to implement and guards you against all data modification races, I think. However it puts some restrictions on the underlying FS implementation.

antoinealb:
> What if the file is updated not through the UAVCAN file server, but by any other external influence? For example, by the local operating system, or some other interface.

Well that would have to be a file-level metadata, not only specific to UAVCAN.

pavel-kirienko:
> Well that would have to be a file-level metadata, not only specific to UAVCAN.

But that would require all users of the file system to adhere to some protocol, would it not? Otherwise, who would ensure the integrity of the metadata? Ideally, we should build a general-purpose API that can be slapped on top of any file system without any extra preconditions.

> This is fine by me, but I have a slight preference to the next one as it doesn't require magical filenames.

I am curious what would @thirtytwobits say.

antoinealb:
> But that would require all users of the file system to adhere to some protocol, would it not?

It would. But that is also the case for a protocol based on file modification time, for example, which would have to adhere to some semantics.

thirtytwobits:
I'm of mixed mind on this. On the one hand I agree with Pavel in that the current API allows for nodes to remain very stateless and simple but on the other hand without some way of ensuring consistency the API might be useless for some use cases.

Perhaps the safest course of action is to release this API without specific consistency mechanisms and a non-normative section discussing conventional strategies for use. We could then accept proposed improvements based on real-world use of the API.

pavel-kirienko:
Okay, I am going to add a weak recommendation to use signature files in the Read service type definition.

@kjetilkjeka please confirm that the file modification time used here will remain as an integral Unix timestamp. We discussed this at the last call and in https://github.com/UAVCAN/dsdl/pull/50 but I don't think a final resolution was made yet.

kjetilkjeka:
> Perhaps the safest course of action is to release this API without specific consistency mechanisms and a non-normative section discussing conventional strategies for use. We could then accept proposed improvements based on real-world use of the API.

As long as we intend (and design the types to allow) adding a consistency mechanism :+1:

> @kjetilkjeka please confirm that the file modification time used here will remain as an integral Unix timestamp. We discussed this at the last call and in #50 but I don't think a final resolution was made yet.

Confirmed :+1:

pavel-kirienko:
@thirtytwobits we need your approval.

Issue #55, opened 2018-11-20
Archive the vendor-specific DSDL repo


pavel-kirienko:
Archive the vendor-specific DSDL repo, update the README file here to indicate that this repository is supposed to contain all public definitions, both standard and vendor-specific.

Repo to be archived: https://github.com/UAVCAN/vendor-specific-dsdl

kjetilkjeka:
A different proposal to how this can be done

https://forum.uavcan.org/t/simpler-and-better-dsdl-versioning/280

PR #56, opened 2018-11-27
Common repository for regulated identifiers (both standard and third-party)

pavel-kirienko:
See https://github.com/UAVCAN/dsdl/issues/55


pavel-kirienko:
Blocked by https://forum.uavcan.org/t/simpler-and-better-dsdl-versioning/280

Issue #63, opened 2019-03-17
Representation of an invalid ambiguous timestamp


pavel-kirienko:
As the ambiguous timestamp value is overflowing, none of its values can be reserved for special purposes; hence, the following is a mistake:

https://github.com/UAVCAN/dsdl/blob/dee68f6a1ef42efb82c56e2811c3c094ce9ea2dd/uavcan/time/SynchronizedAmbiguousTimestamp.1.0.uavcan#L11-L12

We may need to sacrifice one bit to make the value optional, like:

```
uint27 OVERFLOW_PERIOD_us = 83886080
truncated uint23[<=1] decamicrosecond
```

pavel-kirienko:
...which would make it variable-length, breaking alignment wherever it's used. We could do this instead:

- Option A: Make it variable-length as described but move all usages towards the end, so that the alignment of other fields is unaffected. This is not a great option because it restricts usage of this type (what if it can't be moved toward the end?).

- Option B: Add a bit flag alongside indicating whether the value is valid. This is not type safe, not great. We could alternatively use a union, which would be bitwise identical but type safe:

```python
@union
Void23.1.0 empty
truncated uint23 decamicrosecond
```

Where `Void23.1.0` is defined as `void23` (can't use voids as union members directly, at least not yet).

- Option C: Resurrect the old timestamp-related debates ([1](https://forum.uavcan.org/t/si-namespace-design/207/14?u=pavel.kirienko), [2](https://forum.uavcan.org/t/on-timestamping/237)) and see if we can either go back to 56-bit timestamps or find some cleverer solution. I never liked the short timestamp (as explained in the linked topics) but there didn't seem to be any other solution.

thirtytwobits:
will comment on this tomorrow (May 15th)

thirtytwobits:
Option B seems the most reasonable since some uses of this type will not require special values, therefore; B seems to optimize for the simplest use case.

I don't get the union idea though. How does that solve anything?

pavel-kirienko:
Suppose you have a flag like this:

```python
bool valid
truncated uint23 decamicrosecond
```

When you receive a message, you can go about handling it like this:

```c++
if (message.timestamp.valid)  // Danger! Easy to forget.
{
    handle(message.timestamp.decamicrosecond * 10, message.value);
}
```

[...]

PR #67, opened 2019-08-30
v1 finalization

pavel-kirienko:
- Updated the CI test script to enforce a consistent DSDL coding style.
- Enlarged the buffers in data transfer types (file and Internet) to make use of all of the available space.
- Updated the data type documentation to make it transport-agnostic.
- Stabilized key data types.
- Refactored the SI namespace to accommodate #63: now there are two sub-namespaces:
  - `uavcan.si.sample` -- everything from `uavcan.si` is moved there, the overflowing timestamp replaced with non-overflowing version, the former removed completely.
  - `uavcan.si.unit` -- just SI types with no metadata (no timestamps, just values).

I hope the terms `sample` and `unit` make sense; if not, please suggest alternatives.


uavcan/pnp/32741.NodeIDAllocationData.0.1.uavcan
```diff
@@ -180,18 +182,18 @@
...
+# If the allocatee does not have any preference, it should request the highest possible node-ID. Keep in mind that
+# the two highest node ID values are reserved for network maintenance tools; requesting those is not prohibited,
 # but the allocator is recommended to avoid granting these node-ID, using nearest available lower value instead.
 # The allocator will traverse the allocation table starting from the preferred node-ID upward,
 # until a free node-ID is found (or the first ID reserved for network maintenance tools is reached).
 # If a free node-ID could not be found, the allocator will restart the search from the preferred node-ID
 # downward, until a free node-ID is found. In pseudocode:
 #
-#   uint8_t findFreeNodeID(const uint8_t preferred)     // Returns >127 on failure (valid node-ID range from 0 to 127)
+#   uint16_t findFreeNodeID(const uint16_t preferred)   // Returns 0xFFFF on failure
```
thirtytwobits: would a node ID not be an IPv6 address for that transport? I.e. does this need to handle 128 bit addresses?

test.py
```diff
@@ -51,36 +58,53 @@ def compute_max_num_frames_canfd(bit_length):
...
+        # Check header comment
+        if index == 0 and line != '#':
+            if not any(map(lambda e: t.full_namespace.startswith(e),
+                           NAMESPACES_EXEMPTED_FROM_HEADER_COMMENT_REQUIREMENT)):
+                abort('Every data type definition must have a header comment surrounded with "#\\n",',
+                      'unless it is a member of:', NAMESPACES_EXEMPTED_FROM_HEADER_COMMENT_REQUIREMENT)
+
+        # Check trailing comment placement
+        # TODO: this test breaks on string literals containing "#"
+        if not line.startswith('#') and '#' in line and '  #' not in line:
```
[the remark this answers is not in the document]
pavel-kirienko: Yes. But, to do this *really correctly* we need somebody to write a standalone DSDL linter script. I should open a ticket perhaps. This script here is a quick hack.

uavcan/pnp/32741.NodeIDAllocationData.0.1.uavcan
pavel-kirienko: Probably not, but this is very debatable. See https://forum.uavcan.org/t/alternative-transport-protocols/324, search for "determinism considerations".

PR #70, opened 2019-09-13
Resolve the TAI spec ambiguity

pavel-kirienko:
State explicitly that the TAI timestamp represents the number of SI time units (i.e., fixed length) elapsed since the midnight of 1 Jan 1970.

### DIFF
--- a/uavcan/time/TimeSystem.0.1.uavcan
+++ b/uavcan/time/TimeSystem.0.1.uavcan
@@ -8,7 +8,8 @@
 # Monotonic time is a time reference that doesn't change rate or make leaps.
 uint4 MONOTONIC_SINCE_BOOT = 0
 
-# International Atomic Time; https://en.wikipedia.org/wiki/International_Atomic_Time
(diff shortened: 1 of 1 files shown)


uavcan/time/TimeSystem.0.1.uavcan
```diff
@@ -8,7 +8,8 @@
 # Monotonic time is a time reference that doesn't change rate or make leaps.
 uint4 MONOTONIC_SINCE_BOOT = 0
 
-# International Atomic Time; https://en.wikipedia.org/wiki/International_Atomic_Time
+# International Atomic Time; https://en.wikipedia.org/wiki/International_Atomic_Time.
```
thirtytwobits: Wait, so...we have to deploy all new firmware every-time a leap-second is added to the TAI or does GPS also add the leap-second at the same time?

pavel-kirienko: Neither do. The leap second is only considered in UTC. As the rotation of our planet decelerates, UTC and TAI aräe slowly diverging.

Issue #66, opened 2019-07-08
PnP cluster protocol updates


pavel-kirienko:
The log index should be widened from uint8 to match the maximum width of node-ID, which is 16 bit.

https://github.com/UAVCAN/public_regulated_data_types/blob/f468909db282e524eb6410187f02a33720f196d4/uavcan/pnp/cluster/390.AppendEntries.1.0.uavcan#L77-L78

https://github.com/UAVCAN/public_regulated_data_types/blob/f468909db282e524eb6410187f02a33720f196d4/uavcan/pnp/cluster/391.RequestVote.1.0.uavcan#L8

pavel-kirienko:
Fixed in #72

PR #72, opened 2019-10-13
Fix to #66 and minor clarifications

pavel-kirienko:
A review is optional.

After this PR is merged, I propose to swap the branches so that the current master becomes the dedicated v0 branch and the current `uavcan-v1` becomes the new master. This is already done for PyUAVCAN: https://github.com/UAVCAN/pyuavcan/branches

Many data types are not yet stable, but thanks to the new versioning system we don't need to stabilize them in order to stabilize the whole namespace. Thanks, Kjetil.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,6 @@ language: python
 python:
   - "3.6"
 before_install:
-  - pip install pydsdl>=0.7
+  - pip install pydsdl~=1.0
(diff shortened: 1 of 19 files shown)


thirtytwobits:
Change looks good. I'm okay with the switch over but please test any other projects that use this as a submodule.

PR #88, opened 2020-10-20
[DS-015] Air data computer

pavel-kirienko:
### Refactor the physics namespace

It might look like a huge changeset, but it is not. There are no semantic changes, just renamings:

- `reg.drone.{physics, service}` => `reg.drone.{phy, srv}` because we are hitting the name length limit.
- Dynamic and kinematic states have been slightly reorganized with no semantic changes to allow the addition of navigation types soon.

Also, new types: linear velocity with covariance, 1D kinetic state with covariance and timestamp.

### Add the air data computer service

This is just a description with a single constant defining the min data update rate.

***One component is missing here:*** the filter delay parameter is not defined. It should go into a separate type published at a low rate over a separate subject, but before we do that, I need to understand its purpose. Per @TSC21, a copypasta from our Slack:

> that is the maximum time offset between the measurement time and the filter process time
> it is a constant, and depends on the type of sensor and the type of filter
> so it's tricky to setup
> so this is for example the max time offset between a air speed sample time of measurement and the time when it gets processed on an EKF to update the horizontal velocity state

What is the prescribed behavior if a sample taken at `Ts` is received by the filter at `Tr` such that `(Tr-Ts) > filter_delay`? Should the filter discard the sample? If this value is also a function of the type of filter, why is it reported by the sensor?


reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
```diff
@@ -0,0 +1,10 @@
+# This is a structural subtype of LinearTs.
```
TSC21, line 1: It doesn't make sense to name this Cov, because the Covariance is between two different random variables, and in this case you are just passing the variances (which is specific to the random variable).

reg/drone/phy/kinematics/trans/Velocity1CovTs.0.1.uavcan
```diff
@@ -0,0 +1,9 @@
+# Linear velocity with timestamp and covariance.
+# Observe that this is a structural subtype of uavcan.si.sample.velocity.Scalar.1.0.
+# For a non-timestamped estimate without covariance use the raw SI type directly.
+
+uavcan.si.sample.velocity.Scalar.1.0 velocity
+
+float16 error_variance  # [(meter/second)^2]
```
TSC21, line 7: Same thing here about the usage of Cov.

reg/drone/phy/thermodynamics/PressureTempCovTs.0.1.uavcan
```diff
@@ -0,0 +1,15 @@
+# Timestamped fluid pressure and temperature (sampled synchronously) with covariance.
+# Observe that this is a structural subtype of uavcan.si.sample.pressure.Scalar.1.0.
+
+uavcan.time.SynchronizedTimestamp.1.0 timestamp
+
+uavcan.si.unit.pressure.Scalar.1.0    pressure
+uavcan.si.unit.temperature.Scalar.1.0 temperature
+
+float16[3] covariance
```
TSC21, line 9: How does this define the URT? These should be `float16[6]` then.

reg/drone/phy/thermodynamics/PressureTempCovTs.0.1.uavcan
```diff
@@ -0,0 +1,15 @@
...
+# The upper-right triangle of the covariance matrix (following the matrix packing rules defined in Specification).
```
TSC21, line 10: I think you are messing URT with the covariance matrix diagonal (which represents the variances.

| xx | xy | xz |
| ... | yy | yz |
| ... | ... | zz |

An acceptable convention would be using a row-major representation of it. Which results in:
`float16[6] covariance = [xx, xy, xz, yy, yz, zz]`

This usage of the URT for representing the covariances is only acceptable because, as you know, the covariance matrices are symmetric.

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,48 @@
+# A generic air data computer (ADC) service.
```
TSC21, line 1: ADC generates confusion (with analog-digital converter). Please replace this with something like air data sensor (ADS).

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,48 @@
...
+#
+# An air data computer service is generally a non-interactive publish-only service: when activated,
+# it keeps publishing its measurements at a fixed rate over several subjects until deactivated.
+# The activation/deactivation, if supported, is managed via the standard arming command service.
```
TSC21, line 5: Using an arming command here doesn't make much sense to me, and I am sure it will generated confusion between adopters. You usually associate arming to ESC/motors/servos, so actuators, not sensors, at least in the avionics world. We might need a parallel `activation` command service for these specific cases.

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,48 @@
...
+# The data update rates and other parameters are controlled via the Register API using implementation-defined
```
TSC21, line 6: We need to standardize this in some way on the service level (in DS-015). Otherwise it generated ambiguity and divergence.

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,48 @@
...
+# An armed ADC should publish the following data subjects synchronously at the same configurable rate
+# which should not be lower than the specified limit. The measurements should be low-pass filtered
+# to avoid frequency aliasing effects.
+#
+#   PUBLISHED SUBJECT       TYPE                                                NOTE
+#   calibrated airspeed     reg.drone.phy.kinematics.trans.Velocity1CovTs
+#   true airspeed           reg.drone.phy.kinematics.trans.Velocity1CovTs       optional
+#   pressure altitude data  reg.drone.phy.kinematics.trans.LinearCovTs          assume ISA; NED frame (+down, -up)
+#   static air data         reg.drone.phy.thermodynamics.PressureTempCovTs      pressure and OAT
+#
```
TSC21, line 28: We need the air temperature, otherwise this is unusable. Also, sensor temperature should be a must for all sensor types.

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,48 @@
...
+# Observe that there is no subject for indicated airspeed. This is because per the principles of service-oriented
+# design, the ADC is responsible for applying the necessary transformations on its data to render it ready for
+# consumption by clients. If desired, indicated airspeed may be reported over an implementation-defined subject.
```
TSC21, line 31: I am not sure about this. I will request to an SME if this is actually something expected to come out of the sensor or not. I understand we need to follow the service-oriented design, but without proper extension of the service these data types are unusable.

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,48 @@
...
+#
+# In addition to the above, an armed ADC should publish the following service subjects at an implementation-defined
+# rate that should not be lower than the specified limit:
+#
+#   PUBLISHED SUBJECT       TYPE
+#   heartbeat               reg.drone.srv.common.Heartbeat
+#
+# Despite being a non-interactive service, an ADC is recommended to subscribe to the arming command subject.
+# This recommendation may be omitted if the ADC does not support arming state selection, in which case it should
+# always report itself as being ARMED.
```
TSC21, line 41: Again, let's rethink this concept of arm and replace it with activation for sensors.

TSC21 (changes requested):
@pavel-kirienko it generally is good but we need to start thinking of who's going to use the data types and how they expect to use them. We are too much focused on service-oriented designs but then we miss the point of DS-015 which is also support standardized network services that are easy to use without the need to rethink the implementation and force the vendors to extend it in a way that requires both sides (autopilot maintainers and peripheral vendors) to readjust the source code every time.

So my suggestion is that we keep this approach but at the same time, provide the last mile: this means, provide at least standardized data type that composes the usage of the base data types being considered here.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: That is true but the objective here is to follow a consistent convention. We need a generic term that is applicable both to covariance and error variance, I figured that `Cov` is an appropriate choice. Do you have a better idea that would work in both cases?

reg/drone/phy/thermodynamics/PressureTempCovTs.0.1.uavcan
pavel-kirienko: You are missing that the covariance matrix is 2x2 elements, so its URT contains 3 elements. This is correct.

reg/drone/phy/thermodynamics/PressureTempCovTs.0.1.uavcan
pavel-kirienko: Same -- you need to notice that the matrix is 2x2, so the URT contains 3 elements.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: The outside air temperature is included in `reg.drone.phy.thermodynamics.PressureTempCovTs`

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: An air data sensor is a different type of equipment, so it cannot be used here. I have just replaced all uses of ADC with either "air data computer" (unabbreviated) or just "service".

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: Making the right decision on what to standardize is at least as important as deciding what *not* to standardize. Nano-management and overspecialization by the standard may harm the ecosystem just as badly as the lack of clear standards; the recent story with OpenCL v3 appears to be an interesting illustration of that.

Our objective here is to prepare a set of building blocks that can be mixed and matched by the integrator to create distributed systems, where it is up to the integrator to configure the involved elements correctly. The auto-configuration is something that we have on the roadmap past the MVP milestone, and it may or may not involve defining standard register names, but at any rate, right now it is unnecessary and premature.

The v1 of DS-015 will require editing many registers per each involved node, but this is something that we are already agreed on, as we were discussing it for some time already and we already have scheduled the work on the CLI register editing tool to be done in the near future.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: Observe that we discussed the arming service as a generic entity that is applicable beyond actuators in the original PR. Let me add screenshots here for context because links to conversations work poorly on GitHub:

image

image

I think the old PR showed that the concept of arming applied to other types of equipment does not cause confusion. I am keen on keeping some generic means of representing the degree of readiness but I am largely indifferent to its naming. If we perceive that arming is a poor choice of word, we can search for a new one although it's probably not worth the effort.

pavel-kirienko:
@TSC21 can you please elaborate on the filter delay as I requested in the OP comment?

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
TSC21: I don't. I would say that or we propagate the covariances (which increases the data type size), or we just send the variances, which in most of the cases is enough.

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: What I am saying is that the concept of arm is something that is quite intrinsic to what you do in actuators. Not sensors. So the wording IMHO should not be applied the same way for both cases.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: The quantities contained in this message are not correlated, so if we included the covariance it would be a (nearly) diagonal matrix anyway. The question is, can we accept `Cov` as a convention even if we're not really dealing with covariance?

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: > The v1 of DS-015 will require editing many registers per each involved node, but this is something that we are already agreed on, as we were discussing it for some time already and we already have scheduled the work on the CLI register editing tool to be done in the near future.

I think there's a miscommunication here: one thing is that you are certain of which registers need to live in a specific device, which has specific functions associated with it, and the the thing that you edit it's the values of the registers, not the register name by itself (also you don't want to manually created more registers, because it's clear what are the functions that device has). The other is to actually allow the edition of everything, which while still possible, it doesn't facilitate the work of the integrator - having to define everything is a burden and should be avoided.

I am not saying this is something that needs to exist for v1 of DS-015. What I am saying is that we need to consider that most of these things should be easy enough to integrate and expand in a way that are easy to adopt and use. Otherwise, no one is going to bother.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: I am going to move this question to the forum and if necessary we will address it in a later PR.

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: > The outside air temperature is included in reg.drone.phy.thermodynamics.PressureTempCovTs.

So then we go back to the situation of defining macro composed types that define the required set of fields that are required for a specific service to be usable by a drone.

> or it can be simply published on a dedicated subject

That doesn't make sense because you will receive the data at the same time. You don't want the PressureTemp to be received on a certain, and the rest of the data on others. The question here is: how do you make sure that you get all of this data with the same timestamp on a more grouped manner - I would say that we need to define a composed type that answers that.

TSC21:
> What is the prescribed behavior if a sample taken at Ts is received by the filter at Tr such that (Tr-Ts) > filter_delay? Should the filter discard the sample?

I would say yes and it should be clear on the consumer side that that happened (like using an error counter, or discarded samples counter).

> If this value is also a function of the type of filter, why is it reported by the sensor?
It's also a function of the type of the sensor, and it takes more relevance than the type of filter.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: https://forum.uavcan.org/t/service-readiness-level/967

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: Moved to the forum https://forum.uavcan.org/t/service-readiness-level/967

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
TSC21: > The question is, can we accept Cov as a convention even if we're not really dealing with covariance?

Short answer, no. And doing it will only generate confusion. Or you pass the URT of the covariance matrix or you just pass the variances (which will be the diagonal of the covariance matrix).

The other thing I am seeing now: why is `Linear` bidimensional and not tridimensional?

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
```diff
@@ -0,0 +1,10 @@
...
+# Use best guess if the error variance is unknown.
+
+LinearTs.0.1 value
+
+float16 position_error_variance         # [meter^2]
```
TSC21, line 6: how exactly is this just a `float16? You have position variance in x, in z and in z.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
```diff
@@ -0,0 +1,10 @@
...
+float16 velocity_error_variance         # [(meter/second)^2]
```
TSC21, line 7: same thing here

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
```diff
@@ -0,0 +1,10 @@
...
+float16 acceleration_error_variance     # [(meter/second^2)^2]
```
TSC21, line 8: and here

reg/drone/phy/kinematics/trans/Velocity1CovTs.0.1.uavcan
TSC21: This is not right. if you are talking about variance, then the variance needs to be in X, Y and Z.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: The challenge is to define types that are sufficiently rich to include all relevant data, and at the same time do not violate the interface segregation principle. Pressure and temperature are generally tightly coupled, hence we have a dedicated type here, which happens to be a structural subtype of just pressure (so one will be able to subscribe using `uavcan.si.sample.pressure.Scalar`).

Grouping everything into one large type that contains everything is against the design principles we're following here. The rationale was discussed at length earlier so there is probably no point to repeat that. Separation into multiple messages does not mean that one would be unable to process them atomically; this case is  addressed in the Specification:

image

reg/drone/phy/thermodynamics/PressureTempCovTs.0.1.uavcan
TSC21, line 9: Actually, this should be split into `pressure_scalar_variance` and `temperature_scalar_variance`. I don't see why we need a covariance of both.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: This is a one-dimensional state.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: This is a one-dimensional state.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: This is a one-dimensional state.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: > Short answer, no. 

Do you have a specific suggestion?

> The other thing I am seeing now: why is Linear bidimensional and not tridimensional?

It is one-dimensional because it is only used to represent one-dimensional kinetic states. Multidimensional geometry will be supported later under the existing framework.

reg/drone/phy/kinematics/trans/Velocity1CovTs.0.1.uavcan
pavel-kirienko: It is a one-dimensional state.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
TSC21: > It is one-dimensional because it is only used to represent one-dimensional kinetic states. Multidimensional geometry will be supported later under the existing framework.

One dimension means that, for example, for position, you have the x variance OR the y variance OR the z variance.

reg/drone/phy/kinematics/trans/Velocity1CovTs.0.1.uavcan
TSC21: which one? X, Y or Z?

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
pavel-kirienko: That is a very astute observation. Let's have a call, this is going nowhere.

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: @pavel-kirienko are you aware of the complexity of synchronizing samples from different sensors to be fused in a filter? Why are we pushing that complexity here?

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
bresch: I would say that using `cov` won't create any ambiguity: the matrix containing both variances and covariances is called "covariance matrix" and the variance is just a special name when both random variables are the same.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
TSC21: I see now the point. But as we discussed, it doesn't make sense to create a data type that fits one dimension and propagate a variance for one dimension and call it `position_error_variance`. What is conventional (you can check ROS or Mavlink as an example), is that you create a message/data type which is associated with linear/angular position/velocity/acceleration, and those are are 3 dimensional (points, vectors), or 4D if they are stamped.

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
TSC21: answer above

reg/drone/phy/kinematics/trans/LinearCovTs.0.1.uavcan
TSC21: answer above

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: @bresch can I have your comment here?

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: Is there any action item specific to this pull request that you would like to propose?

reg/drone/phy/thermodynamics/PressureTempCovTs.0.1.uavcan
pavel-kirienko: Pressure and temperature are tightly correlated, so my assumption is that this information could in theory be leveraged by an external state estimator. The cost of managing a single extra `float16` is low, so I added it. This case is different from, say, kinematic states, because they are known to be uncorrelated.

reg/drone/srv/air_data_computer/_.0.1.uavcan
bresch: My question is "what do we expect from a sensor that cannot compute CAS"? Should it simply publish IAS in the CAS field? How the client would then know that it should apply calibration on that data?

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: I think we need to make sure we standardize register naming that correspond to things that are used basically in all nodes for most of the topics, like the update rate, and also the registers for pub/sub port IDs. This will make sure that we have a strong convention that guarantees that the integrator knows what registers to look for when it wants to change the register values.

So for this particular PR, I would start by defining what registers should by default exist on a node with a particular function and its naming. Other registers can exist, but as you specify already, those can be implementation-defined.

reg/drone/srv/air_data_computer/_.0.1.uavcan
```diff
@@ -0,0 +1,50 @@
...
+# to avoid frequency aliasing effects.
+#
+#   PUBLISHED SUBJECT       TYPE                                                NOTE
+#   calibrated airspeed     reg.drone.phy.kinematics.trans.Velocity1VarTs
+#   true airspeed           reg.drone.phy.kinematics.trans.Velocity1VarTs       optional
+#   pressure altitude data  reg.drone.phy.kinematics.trans.LinearVarTs          assume ISA; NED frame (+down, -up)
+#   static air data         reg.drone.phy.thermodynamics.PressureTempVarTs      pressure and OAT
+#
+# Observe that there is no subject for indicated airspeed. This is because per the principles of service-oriented
+# design, the air data computer is responsible for applying the necessary transformations on its data to render it
```
sfuhrer, line 30: I'm mission some context here. Is the air data computer meant as a "sensor", meaning that if you have multiple airspeed sensors you also get multiple publications? Or is more of an "estimator", that fuses multiple airspeed readings and publishes an averaged estimate/the best instance?
Depending on the above, if it's a "sensor" message then I would rather make it publish IAS istead of CAS, and leave the calibration to the estimator. And consequently also leave away TAS, as it can't be estimated without knowing CAS. If it's an "estimator" I think publishing CAS and TAS is fine.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: > My question is "what do we expect from a sensor that cannot compute CAS"? Should it simply publish IAS in the CAS field? How the client would then know that it should apply calibration on that data?

The short answer is that if a sensor is unable to produce highly processed data that is abstracted away from the means of its estimation, it is non-conformant. This is a very conscious requirement; we need to keep in mind that this standard is intended to enable distributed intravehicular systems and it is not a bad clone of I2C. The long answer is provided in [The Guide](uavcan.org/guide/).

The vendor of a [COTS air data computer](https://www.tindie.com/products/avionicsanonymous/uavcan-air-data-computer-airspeed-sensor/) would provide configuration parameters (see Register API) that define the relationship between IAS and CAS. The parameters are to be configured by the integrator based on the position of the pitot sensor and other relevant properties.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: There may be multiple network service instances per node, so I don't think it's possible to define specific register names. We could define wildcards as has been proposed for `uavcan.sub.*`, for example. But notice that for a human integrator it doesn't make much difference whether, say, the rate of a particular sensor feed is set via a spec-adherent register or some implementation-defined register. Where it matters is the auto-configuration, but this capability is expressly excluded from the scope of DS-015 v1.0. My plan is to leave this undefined for now, and get back to it after we have the MVP out.

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: > The short answer is that if a sensor is unable to produce highly processed data that is abstracted away from the means of its estimation, it is non-conformant. This is a very conscious requirement; we need to keep in mind that this standard is intended to enable distributed intravehicular systems and it is not a bad clone of I2C.

@pavel-kirienko so we just assume that a vendor needs to build a sensor which in the end will be much more expensive because it can provide CAS? I mean, I don't think this is a reasonable way of seeing things. DS-015 is aimed for a board spec of solutions, from the less to the most expensive. It's more expensive to create a COTS air computer like that than providing a field for CAS.

pavel-kirienko: IAS-CAS translation does not require any additional hardware, but a trivial piecewise transfer function, so I don't think the cost is affected. If the cost is really critical, I2C is the way to go, otherwise you're going against the core design goals of the standard.

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: > We could define wildcards as has been proposed for uavcan.sub.*, for example

Let's do that as an intermediate step here then, please.

pavel-kirienko: Not possible without redefining the project goals and disrupting the schedule. The plan is to release the standard in October; defining the register names would require exploring the auto-configuration options, it is major work and it will take time.

TSC21: > defining the register names would require exploring the auto-configuration options

Why is that? What does it take to define wildcards on the data types definitions that can define the way that the registers are named? I don't see where exactly does that impact on the project and where does that imply auto-configuration options. We are just stating that a human integrator will know exactly what would be the wildcard to look for when configuring a register. That for me is quite clear.

reg/drone/srv/air_data_computer/_.0.1.uavcan
TSC21: > otherwise you're going against the core design goals of the standard.

What exactly are we going against? One of the core design goals of the standard is to allow flexibility for the HW vendor to specify what sort of data it wants to provide.

pavel-kirienko: The goals are defined in the MVP doc in the form of a reference to the interface design guidelines. The values announced there are, therefore, the core values of DS-015. We are venturing too far away from the scope of this PR here with this discussion.

> The main motivation for our (Zubax) involvement in this project is that we perceive that the early draft of the standard proposed for public evaluation is not built according to the best practices of diligent design and it is not an adequate candidate for a long-term industry standard. Given the strategic importance of the standard both for the PX4 ecosystem and for the core UAVCAN project itself, leaving the project on its current trajectory may reflect poorly on the future of the entire industry. Our criticism of the proposed design is expressed in the form of advisory guidance in the freshly published chapter on Interface Design Guidelines. Despite the design guidance being available, it does not appear to be sufficient to steer the design process off its current trajectory, so we choose to intervene directly.
>
> To implement the work described in this document we request authority to override design decisions made by the Auterion team.

reg/drone/srv/air_data_computer/_.0.1.uavcan
pavel-kirienko: > What does it take to define wildcards on the data types definitions that can define the way that the registers are named?

It depends on what our final intent is.

If the intent is to define some arbitrary requirements, it can be done in a minute.

If the intent is to define a self-consistent and viable standard, it takes work. The register name patterns are an inherent part of the auto-configuration capability, and it will be impossible to define them without constructing the general framework of the auto-configuration service first.

What you are requesting is impossible to do under the current schedule and budget.

pavel-kirienko:
@TSC21 FYI, remaining items that are to be completed before UAVCAN v1.0 stable is out: https://github.com/UAVCAN/public_regulated_data_types/milestone/1

Of those, the following block DS-015 MVP: https://github.com/UAVCAN/public_regulated_data_types/issues?q=is%3Aopen+is%3Aissue+label%3Areg-drone

TSC21 (approved):

Issue #90, opened 2020-10-22
[DS-015] Rename "Arming"


pavel-kirienko:
Rename `reg.drone.srv.common.Arming` and its members into something more generic, leaving the semantics unchanged.

Context: https://forum.uavcan.org/t/service-readiness-level/967

PR #91, opened 2020-10-24
[DS-015] Minor non-semantic adjustments

pavel-kirienko:
Fixes https://github.com/UAVCAN/public_regulated_data_types/issues/90

* Introduce minor clarifications and avoid overspecification.
* Propose fix to #90 by renaming without changing the semantics:
  * Arming --> Readiness
  * DISARMED --> STANDBY
  * ARMED --> ENGAGED

### DIFF
--- a/reg/drone/README.md
+++ b/reg/drone/README.md
@@ -93,15 +93,15 @@ UAVCAN implementation libraries are optimized for handling multi-frame transfers
 
 ## Conventions
 
-All physical quantities except error variance should be represented as `float32` by default.
-Error variance and covariance matrices should use `float16` by default.
+- All physical quantities except error variance should be represented as `float32` by default.
+  Error variance and covariance matrices should use `float16` by default.
 
-Covariance matrices should be represented as their upper-right triangles using the matrix packing rules
-defined in the Specification.
+- Covariance matrices should be represented as their upper-right triangles using the matrix packing rules
+  defined in the Specification.
 
-Types with (co)variance should be suffixed `Var`; types with timestamp should be suffixed `Ts`;
-types with both should be suffixed `VarTs`.
-The timestamp field, if present, should be the first one;
-error (co)variance information should follow the data field(s) it relates to.
+- Types with (co)variance should be suffixed `Var`; types with timestamp should be suffixed `Ts`;
+  types with both should be suffixed `VarTs`.
+  The timestamp field, if present, should be the first one;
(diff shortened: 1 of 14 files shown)


reg/drone/README.md
```diff
@@ -103,3 +103,5 @@ Types with (co)variance should be suffixed `Var`; types with timestamp should be
 types with both should be suffixed `VarTs`.
 The timestamp field, if present, should be the first one;
 error (co)variance information should follow the data field(s) it relates to.
+
+Publishers of measurements or estimates should apply low-pass filtering to avoid frequency aliasing.
```
TSC21, line 107:
```suggestion
* Publishers of measurements or estimates should apply low-pass filtering to avoid frequency aliasing.
```
Please do the same to the previous. We should aim for bullet points and simplicity always when possible.

reg/drone/phy/acoustics/Note.0.1.uavcan
```diff
@@ -1,19 +1,9 @@
 # Description of a generic musical note in terms of basic physical quantities.
-# This type may be used to control sound notification emitters.
+#
+# This type may be used to control sound notification emitters assuming the best effort policy:
+# if the requested parameters exceed the capabilities of the emitter, the closest possible values should be assumed.
 
 uavcan.si.unit.frequency.Scalar.1.0 frequency
-# If the requested frequency is outside of the supported range, the command should be ignored.
```
TSC21, line 5: Why are we removing this?

TSC21 (commented):
I like the concept of "Readiness" better yes. And the replacement of armed with engaged also sounds better.

@PetervdPerk-NXP any inputs?

reg/drone/README.md
pavel-kirienko: I think bullet points are often overused when the author should have done a better job writing coherent text, but in this case, they are actually helpful.

reg/drone/phy/acoustics/Note.0.1.uavcan
pavel-kirienko: Because this is a case of overspecification: the generic type intended for modeling sound does not need to define service contracts for sound emitters. If necessary, this will need to be defined externally.

PetervdPerk-NXP:
LGTM, Like I posted in the forum, using another definition then arming/disarming is the way to go.

TSC21 (approved):

PR #92, opened 2020-10-24
Extend the set of standard register names

pavel-kirienko:
Fixes https://github.com/UAVCAN/public_regulated_data_types/issues/89

This will need to be covered in The Guide later in detail.

I am not sure if we should revisit the 50 symbol limit on the name length (registers and data types).


uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,60 @@
...
+#
+#   uavcan\.cookie                                    string          mutable, persistent     (empty)
+#
+# Arbitrary data stored by an external autoconfiguration authority (e.g., a mission computer, an auto-configuration
+# utility, etc.). This data is never to be used by the node itself; instead, the node merely provides raw data
+# storage space.
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan\.pub\.[a-zA-Z_][a-zA-Z0-9_]{0,34}\.id      natural16       mutable, persistent     65535 (unset, invalid)
```
TSC21, line 95: This is not very human-readable, is it?

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,60 @@
...
+#   uavcan\.sub\.[a-zA-Z_][a-zA-Z0-9_]{0,34}\.id      ...             ...                     ...
+#   uavcan\.cln\.[a-zA-Z_][a-zA-Z0-9_]{0,34}\.id      ...             ...                     ...
+#   uavcan\.srv\.[a-zA-Z_][a-zA-Z0-9_]{0,34}\.id      ...             ...                     ...
+#
+# Publication/subscription/client/server port-ID, respectively. These registers are configured by the system integrator
```
TSC21, line 100: We should also add a note that these can be autoconfigurable.

TSC21 (commented):
@pavel-kirienko where's the relationship with functions?

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: Maybe, but this is the POSIX ERE standard.

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: Done.

pavel-kirienko:
> where's the relationship with functions?

This is supposed to be addressed by the following remark:

> For example, a register named "uavcan.pub.measurement.id" defines the port-ID of a measurement published by this node.

TSC21:
> > where's the relationship with functions?
> 
> This is supposed to be addressed by the following remark:
> 
> > For example, a register named "uavcan.pub.measurement.id" defines the port-ID of a measurement published by this node.

I think the naming "function" should be present somewhere in someway,

uavcan/register/384.Access.1.0.uavcan
TSC21: So what is supposed to interpret this syntax? Isn't this user-facing docs?

TSC21 (approved):
LGTM! Thanks @pavel-kirienko!

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,68 @@
...
-#                       |       |           | automatic node-ID. Invalid values other than 65535 should be avoided for
-#                       |       |           | consistency. The factory-default value should be 65535.
-# ----------------------+-------+-----------+--------------------------------------------------------------------------
+# common functions. These conventions are not mandatory to follow, but implementers are recommended to adhere because
+# they enable enhanced introspection capabilities and simplify device configuration and diagnostics.
+#
+#   REGISTER NAME PATTERN                             TYPE            FLAGS                   RECOMMENDED DEFAULT
+# =====================================================================================================================
+#
+#   uavcan.node.id                                      natural16       mutable, persistent     65535 (unset/PnP)
```
TSC21, line 73: Maybe you want to push the cells a bit to the left?

pavel-kirienko: No.

TSC21:
> I am not sure if we should revisit the 50 symbol limit on the name length (registers and data types).

It might be a good idea to increase this.

uavcan/register/384.Access.1.0.uavcan
thirtytwobits, line 73: why natural? Why not Integer16?

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,68 @@
...
+#
+# Contains the node-ID of the local node. Values above the maximum valid node-ID for the current transport
+# indicate that the node-ID is not set; if plug-and-play is supported, it will be used by the node to obtain an
+# automatic node-ID. Invalid values other than 65535 should be avoided for consistency.
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.node.description                             string          mutable, persistent     (empty)
+#
+# Human-readable verbose description of this specific node defined by the integrator (not vendor).
```
thirtytwobits, line 83:
```suggestion
# User-defined, human-readable description of this specific node. This is intended for use by a system integrator and should not be set by the manufacturer of a component. For example: on a quad-rotor drone this might read "motor 2" for one of the ESC nodes.
```

I'm not sure I'd call this verbose with only 256 bytes to work with (especially if you are using a lot of non-ascii characters).

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,68 @@
...
+#   uavcan.cookie                                       string          mutable, persistent     (empty)
+#
+# Arbitrary data stored by an external autoconfiguration authority (e.g., a mission computer, an auto-configuration
+# utility, etc.). This data is never to be used by the node itself; instead, the node merely provides raw data
+# storage space.
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.pub.PORT_NAME.id                             natural16       mutable, persistent     65535 (unset, invalid)
+#   uavcan.sub.PORT_NAME.id                             ...             ...                     ...
```
thirtytwobits, line 96: seeing as [the ditto mark](https://en.wikipedia.org/wiki/Ditto_mark) goes back to 934BC we probably shouldn't try to change it now.

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,68 @@
...
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.node.description                             string          mutable, persistent     (empty)
+#
+# Human-readable verbose description of this specific node defined by the integrator (not vendor).
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.cookie                                       string          mutable, persistent     (empty)
```
thirtytwobits, line 87: as the name suggests, this concept risks becoming a malware enabler depending on how systems decide to use, or abuse, this storage facility. Why do we need this?

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,68 @@
...
+#
+#   uavcan.pub.PORT_NAME.id                             natural16       mutable, persistent     65535 (unset, invalid)
+#   uavcan.sub.PORT_NAME.id                             ...             ...                     ...
+#   uavcan.cln.PORT_NAME.id                             ...             ...                     ...
+#   uavcan.srv.PORT_NAME.id                             ...             ...                     ...
+#
+# Publication/subscription/client/server port-ID, respectively. These registers are configured by the system integrator
+# or an autoconfiguration authority when the node is first connected to a network.
+#
+# The "PORT_NAME" defines the human-friendly name of the port, which is related to the corresponding function
```
thirtytwobits, line 103: "the ... name of the port..." and the user asks, "where do I find this name?"

thirtytwobits (changes requested):

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: Because a node-ID (or subject-ID) cannot be negative.

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: > this concept risks becoming a malware enabler

In the same way as a kitchen knife is a cold weapon, yes. The very same thing can be said about the bootloader protocol, the LAN forwarding interface, and just about every other feature of UAVCAN.

There is currently no cybersecurity measures, and I am not exactly clear on how this side of the project will be evolving in the future. Conventional aerospace systems ensure (cyber)security by preventing unauthorized access to the aircraft's systems; this model may no longer be applicable in the future due to certain new capabilities of modern systems but it's a different story. In my humble opinion, the overall perception of the importance of cybersec in aerospace is somewhat artificially inflated mostly because people experienced in ICT mistakenly carry over core assumptions into a different field where they are no longer valid. This occasionally leads to [appalingly useless security research publications](https://www.rapid7.com/research/report/investigating-can-bus-network-integrity-in-avionics-systems/) that can be effectively summarized as "our new research shows that removing the wing breaks the airplane". At the moment, I assign a low priority to this matter, but it doesn't mean that it is out of the scope entirely or that it is not something that will be worked on. There are far more critical issues out there.

image

Currently, cybersec is backlogged in DS-015 with a note that it may be addressed either in DS-015 itself or in the core UAVCAN Specification, depending on how things develop.

Speaking about this cookie register specifically: it will be relied on by the upcoming autoconfiguration algorithm; it is based on this post in an early revision of the DS-015 document:

> "PX4 node gets the heartbeat..." this requires storing persistent data in a 'node id table'. Storage means we have to worry about maximum sizes of tables (i.e. how much non-volatile memory must your hardware provide to host this function) and consistency (i.e. how do you know the data hasn't become corrupt).
> 
> Another strategy is to require each node to remember its own "I am configured" state. This provides a bounded storage problem distributed to each node. It's also easier for the node to monitor consistency and if there is a failure of its persistent storage you don't lose the entire database; just that one node becomes unconfigured. Perhaps this activity can be redesigned to use some sort of "system ID" where the function can be distributed (Pavel loves distributed!). In such a scheme there is a distribution on the bus of "this is the system ID you are attached to" and if a Node sees it hasn't configured for that system it can enter a system configuration activity. This activity might be the node looking at it's registers and for each that doesn't have a port identifier received from this system it broadcasts "I need to know about port ID X for system Y". The code cannot become operational until all port identifiers it requires for its primary function have been provided to it. It would seem this allows attached devices to better discover and adapt to system capabilities. For example, if I ask for the port identifier for function "myfunction_2_0" and nobody knows I can then ask "does anybody have port identifiers for myfunction_1_0?," etc.
> 
> In this scheme the prompt to a user for "there's an unknown node" would be a function that simply monitored for any node that did not report it was configured for the current system.
> 
> So, this is just some quick thoughts here. Sorry I don't have time to flush them out but in-general I think you need to find a way to remove the central "configured nodes" database here that could be a single point of failure for the system.

This change request is a bit sudden and it is not backed up by a comprehensive proposal. I am cutting corners to ensure that we will be able to retrofit autoconfiguration in DS-015 v1.x (x > 0) without the need to ever touch peripheral node implementations that conform to DS-015 v1.0. If there are objections against such (low-risk) deviation from our normal design process, we can pull this specific register from the changeset, but keep in mind that requiring implementers of DS-015 v1.0 to change the firmware to support autoconfiguration may do damage to the ecosystem.

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: Added a paragraph below this one to add clarity.

uavcan/register/384.Access.1.0.uavcan
thirtytwobits: my comment from the forum was suggesting a read-only register for "I am configured" though. My question is, why is this an externally _writable_ register and not an immutable, volatile register?

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -64,19 +64,73 @@
...
+# or a network service supported by the node. The name shall match the following POSIX ERE expression:
+#
+#   [a-zA-Z_][a-zA-Z0-9_]{0,34}
+#
+# The names are defined by the vendor of the node. The user/integrator is expected to uderstand their meaning and
```
thirtytwobits, line 110:
```suggestion
# The names are defined by the vendor of the node. The user/integrator is expected to understand their meaning and
```

pavel-kirienko: My spell checker is malfunctioning again.

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: The register is expected to store a device-specific context that is entirely determined by the *autoconfiguration authority* that writes and interprets its content. The objective is to relieve said authority from mapping the node identity with its specific settings internally. If you want to discuss this at length, I think it is best to pull this change from this PR and accept the rest, postponing this issue until the point where the DS-015 autoconfiguration proposal is ready.

thirtytwobits: I'm not vehemently opposed to this but I'm not able to really review it since I don't understand the context. It might be best to pull it and reintroduce when that context is available.

Issue #68, opened 2019-09-09
DSDL linter


pavel-kirienko:
We need to have this script transformed into a standalone reusable DSDL linter:

https://github.com/UAVCAN/public_regulated_data_types/blob/uavcan-v1.0/test.py

Business requirements:

- Implemented as a standalone Python script dependent only on PyDSDL.
- Ensure that the formatting recommendations provided in the specification are followed.
- Support for definition-wide diagnostic suppression using pragmas.
- Ensure the header comment is provided for all data types. Allow the caller to specify which namespaces should be allowed to bypass this requirement.
- Allow the caller to specify the maximum size of serialized representations. Provide a pragma to override this.
- Ensure ASCII charset; prohibit tabs and `\r` (carriage return).

PR #93, opened 2020-10-25
Update the port introspection service per discussion on the forum

pavel-kirienko:
Fixes https://github.com/UAVCAN/public_regulated_data_types/issues/84

The new port introspection service follows the push model, simplifying its support in hard real-time systems with strict scheduling constraints (the amount of time required to manage the service is known which is normally not the case for service servers). See https://forum.uavcan.org/t/alternative-transport-protocols-in-uavcan/324/34?u=pavel.kirienko

Contrary to the discussion on the forum:

> The publisher configuration can be subjected to the same treatment but it is not immediately required by the application-level objectives at hand, and as such, this part of the problem can be postponed indefinitely – after all, the publication set is always trivially observable on the network by means of mere packet monitoring, which is not the case for the subscription configuration.

Notice:

- A subject may be published ad-hoc or with a very low interval, which can make its existence unobservable over a finite time interval.
- Some packets may not be visible by the monitor/debugger node due to static routing or acceptance filtering.
- Logical coupling: network introspection is coupled with routing/acceptance filtering.

Which is why I also included the publisher configuration into this service.

The service GetInfo is removed because for fixed-ID ports the data type is obviously deducible from the port-ID, and for non-fixed-ID ports this is addressed in https://github.com/UAVCAN/public_regulated_data_types/pull/92.

The statistics service had an issue with request/response handling; instead of fixing I removed it. We may revisit it later and consider using standard register names instead of a dedicated service to lower the implementation barriers.


uavcan/node/port/433.GetStatistics.0.1.uavcan
```diff
@@ -1,30 +0,0 @@
-# This service is used to obtain statistical metrics of a port (either message or service).
```
thirtytwobits, line 1: You said this had an error with "request/response handling" that led you to remove this. Can you elaborate please?

uavcan/node/port/7510.List.0.1.uavcan
```diff
@@ -0,0 +1,22 @@
...
+# - Subjects that this node is subscribed to (a datalogger or a debugger would typically subscribe to all subjects).
+# - RPC services consumed by this node (i.e., service clients).
+# - RPC services provided by this node (i.e., service servers).
+#
+# All nodes should implement this capability to provide network introspection and diagnostic capabilities.
+# This message should be published using the fixed subject-ID as follows:
+# - At the OPTIONAL priority level at least every MAX_PUBLICATION_PERIOD seconds.
+# - At the OPTIONAL or SLOW priority level within MAX_PUBLICATION_PERIOD after the port configuration is changed.
+#
+# pragma: no-bit-length-limit
```
thirtytwobits, line 12: ?

uavcan/node/port/7510.List.0.1.uavcan
```diff
@@ -0,0 +1,22 @@
...
+
+uint8 MAX_PUBLICATION_PERIOD = 10  # [seconds]
```
thirtytwobits, line 14: what if we suggested a register value for this where 0xFF disables the broadcast, 0 means "as fast as possible", and other values allow this to be adjustable. This would be useful for debugging where you want to up the broadcast rate so you don't have tow wait around for 10 seconds to observe changes in a tool like Yukon.

thirtytwobits (changes requested):

uavcan/node/port/433.GetStatistics.0.1.uavcan
pavel-kirienko: Suppose there is a node that provides a service under some service-ID, and also has a client on the same service-ID. When you send a request with a given service-ID, you can't specify if you are referring to the client or the server.

uavcan/node/port/7510.List.0.1.uavcan
pavel-kirienko: Remember this quick-and-dirty `test.py` DSDL linter we use for validating the data type definitions? We have a tracking issue here documenting the need to create a *proper solution* instead of the current *quick hack*: #68.

The *hack* has this pragma to not fail the check if the data type is too large:

https://github.com/UAVCAN/public_regulated_data_types/blob/2eea7e6bbcc654b1516f6fb3435b8a2d44baaa1b/test.py#L113

If you think it causes confusion, we can easily just drop this check. Its value is rather low anyway.

uavcan/node/port/7510.List.0.1.uavcan
pavel-kirienko: image

The reasoning here is that a system that can easily implement a behavior like that (a low-SIL system, likely) is allowed to publish a new message on change immediately, so in this case, you won't have to wait around. Also, subscription configurations are generally not changed often, even in the conventional ROS/DDS world (this is my humble empirical observation). At the same time, high-SIL systems are allowed to opt-out from the variable-rate publishing (which may be hard to reconcile with hard real-time scheduling) while staying conformant.

thirtytwobits (approved):

PR #94, opened 2020-10-28
Enlarge arrays as discussed

pavel-kirienko:
This is a follow-up to the previous discussion we had here and at the call. This change is not expected to introduce incompatibilities.

The new data types here are actually new versions of existing ones, where the old ones are marked deprecated.

The register API is updated without introducing new versions to avoid changing the fixed service-ID. This is wire-compatible provided that long register names are not used; if they are, requests will fail with deserialization error.

### DIFF
--- a/test.py
+++ b/test.py
@@ -104,12 +104,3 @@ def get_max_bit_length(ty) -> int:
         return max(map(get_max_bit_length, [ty.request_type, ty.response_type]))
     else:
         return max(ty.bit_length_set)
-
-
-for t in output:
-    max_bit_length = get_max_bit_length(t)
-    if max_bit_length > MAX_SERIALIZED_BIT_LENGTH:
-        text = open(t.source_file_path).read()
-        if '#pragma:no-bit-length-limit' not in text.replace(' ', ''):
-            print('The data type', t, 'exceeds the bit length limit of', MAX_SERIALIZED_BIT_LENGTH, file=sys.stderr)
-            sys.exit(1)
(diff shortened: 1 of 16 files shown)


TSC21:
Do you actually feel there's a real need to deprecate the data types instead of updating the existent ones? I mean are there any actual systems in production using these data types? I think this will just introduce undesirable noise in the repo.

pavel-kirienko:
I think we should get used to it because any practical namespace will contain legacy data types.

TSC21:
> I think we should get used to it because any practical namespace will contain legacy data types.

Yes but don't you feel this is a little to early for that?

pavel-kirienko:
> Yes but don't you feel this is a little to early for that?

We are almost ***stable*** so it's time to start deploying the backward compatibility measures we have designed. This is what our adopters would expect: the spec has not yet reached the release, but things are already stable enough to (carefully) rely on.

thirtytwobits (approved):

TSC21:
> > Yes but don't you feel this is a little to early for that?
> 
> We are almost _**stable**_ so it's time to start deploying the backward compatibility measures we have designed. This is what our adopters would expect: the spec has not yet reached the release, but things are already stable enough to (carefully) rely on.

Ok. In the other hand, should we consider creating a `depreacted` namespace to where to move these, or that would break compatibility?

pavel-kirienko:
> In the other hand, should we consider creating a deprecated namespace to where to move these, or that would break compatibility?

We want to keep deprecated data types mostly untouched to avoid breaking third-party code that relies on them, so they shouldn't be moved. Deprecated types eventually go to Valhalla anyway so the noise is bounded.

PR #97, opened 2020-12-09
GNSS Sensor Device

aentinger:
This is a very first step into describing the UAVCAN V1 messages for a generic GNSS location sensor device. My idea is to discuss within this PR and develop this message set over various commits. Upon reaching agreement we can "Squash & Merge" the whole PR into a single commit. I don't see another way of doing this as I'm not knowledgeable enough to encompass everyone's requirements. Let's get this started!

The first question I've got is how to actually represent the location data:
* Latitude, longitude, altitude, geoidal separation? If so what units would be best to represent latitude and longitude? `uavcan.si.unit.length.Scalar.1.0` could be used for both altitude and geoidal separation, however there's no unit representing degrees.
* x,y using a cartesian coordinate system such as the [UTM](https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system) reference system? z via altitude and geoidal separation as in the previous suggestion?

CC @pavel-kirienko @thirtytwobits @TSC21 

This fixes https://github.com/107-systems/UAVCAN-GNSS-Node/issues/2 .

### DIFF
--- a/reg/drone/srv/sensor/gnss/Location.0.1.uavcan
+++ b/reg/drone/srv/sensor/gnss/Location.0.1.uavcan
@@ -1,0 +1,5 @@
+# This message type provides location information as obtained via a GNSS sensor.
+
+Source.0.1 source
+
+@sealed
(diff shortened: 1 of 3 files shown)


pavel-kirienko:
Superseded by #100

PR #100, opened 2021-01-08
Extend reg.drone.physics.kinematics with 3D geometry and define the GNSS service

pavel-kirienko:
This PR incorporates and supersedes #97 and unblocks https://github.com/107-systems/UAVCAN-GNSS-Node/issues/2. @aentinger please review.

Start here:

- `reg/drone/physics/kinematics/README.md`
- `reg/drone/service/gnss/_.0.1.uavcan`


reg/drone/service/gnss/FixStatus.0.1.uavcan
```diff
@@ -0,0 +1,13 @@
+# Status of the GNSS navigation solution. This information is needed for indication and general status monitoring only.
+# Data consumers should not use this info and rely on the error estimates reported via covariance matrices instead.
+# If the sensor is equipped with other means of deriving the navigation solution (dead reckoning, VIO, UWB, etc.),
+# then this value may not be representative of the actual status of the solution (only matrices are).
+
+uint2 NONE = 0
+uint2 TIME = 1  # Only time is available.
+uint2 D2   = 2  # The time, latitude, longitude, and NE-plane velocity are available.
```
aentinger, line 8: Is there any specific reason for having this state abbreviated by `D2`? My issue with this shorthand is that I can't infer what it actually means and would rather have it spelled out in long form, i.e. `TIME_POSITION_VELOCITY`. In a similar way `D3` could be `ALL` (or an enumeration as `TIME_POSITION_VELOCITY_HEADING_...`).

reg/drone/service/gnss/Heartbeat.0.1.uavcan
```diff
@@ -0,0 +1,16 @@
...
+Sources.0.1 sources
+
+DilutionOfPrecision.0.1 dop
+
+FixStatus.0.1 fix_status
+
+uint8 num_visible_satellites  # The number of space vehicles visible when the solution was computed.
+uint8 num_used_satellites     # The number of space vehicles whose signals have been utilized to derive the solution.
+
+bool rtk_fix    # True if the current mode is RTK and an RTK fix has been established. Also see Sources.
```
aentinger, line 14: I am curious to know why `rtk_fix` is singled out here and not part of `FixStatus.0.1`?

reg/drone/service/gnss/Sources.0.1.uavcan
```diff
@@ -0,0 +1,37 @@
...
+bool beidou_b1          # 1531.098 MHz
+bool beidou_b2          # 1207.140 MHz
+
+bool sbas               # Satellite-Based Augmentation System (WAAS, EGNOS, GAGAN, QZSS, etc.)
+bool gbas               # Ground-Based Augmentation System, DGPS (IMES, etc.)
+
+bool rtk_base           # Real-Time Kinematics base station data
+
+bool imu                # Inertial Measurement Unit
+bool vo                 # Visual Odometry
```
aentinger, line 27: As I am a fervent believer in spelling out acronyms in code (especially those that can't be easily inferred) I suggest to replace `vo` with `visual_odometry`.

reg/drone/service/gnss/FixStatus.0.1.uavcan
pavel-kirienko: It is supposed to say "2D" and "3D", but [a DSDL identifier cannot start with a digit](https://github.com/UAVCAN/specification/blob/7cacf58936a494de973c81bfe13f1ecbcb71337c/specification/dsdl/grammar.parsimonious#L7). We can change it as you suggested, but upon further consideration, I am actually inclined to replace this with a simple boolean flag indicating whether the fix is attained at all. The available data can be determined from covariance matrices. This change would remove one type from the service definition making it (marginally) simpler.

reg/drone/service/gnss/Heartbeat.0.1.uavcan
pavel-kirienko: The fix status is intended to be an enumeration so I moved it into a dedicated type. As I just wrote above, the current proposal might be over-engineered slightly because the FixStatus can be replaced with a boolean flag next to `rtk_fix` without reducing the utility of the GNSS service. Unless you object, I would add `bool fix` next to this flag instead.

reg/drone/service/gnss/Sources.0.1.uavcan
pavel-kirienko: This is a good catch because it actually goes against the recommendations given in the Spec [section 3.9.1]:

>  Short names, unnecessary abbreviations, and uncommon acronyms should be avoided

reg/drone/service/gnss/FixStatus.0.1.uavcan
aentinger: I agree with your change in 669d7166be579fb8d092ef91caf050f47b667160 ✔️, it is probably a sisyphean task to try to find `FixStatus` enumeration values to suit the requirements of all available GNSS sensors. A boolean flag + already existing covariance matrices should achieve the same result while eliminating redundant information.

reg/drone/service/gnss/Heartbeat.0.1.uavcan
aentinger: Sounds good to me 👍 I'm still wondering why `rtk_fix` requires it's own flag in addition to `fix`? If `fix` is set and `rtk_base` is set, wouldn't that be the same as setting `fix,` `rtk_fix`, and `rtk_base`?

pavel-kirienko: - `fix` indicates that a standalone navigation solution is available.
- `rtk_fix` indicates that an RTK fix is established and the integer wavelength ambiguity is resolved.
- `sources.rtk_base` indicates that base station data (usually encapsulated into an RTCM stream) is available to the receiver, but it says nothing about the status of the navigation solution.

pavel-kirienko:
I just noticed that I forgot to rebase this changeset back to master after the dependent branch was merged in #98. Fixed now.

reg/drone/service/gnss/Heartbeat.0.1.uavcan
aentinger: Thank you for the clarification :+1:

pavel-kirienko:
@aentinger can you please either approve this or introduce further suggestions?

aentinger (approved):

PR #109, opened 2021-02-28
Doc amendment for uavcan.register.Access: mapping between registers and environment variables

pavel-kirienko:
Introduce well-defined mapping between UAVCAN registers and environment variables. Given registers `uavcan.node.id` and `motor.inductance_dq`, one will be able to assign them values 42 and (0.14, 0.16) respectively by exporting environment variables `UAVCAN__NODE__ID=42` and `MOTOR__INDUCTANCE_DQ=0.14 0.16`.

This proposal is motivated by the recent experience with early v1 deployments and my own work on the Yukon backend rewrite (which itself is a UAVCAN-based system). Also, there is a weakly related thread about the possible role of UAVCAN as an IPC transport: https://forum.uavcan.org/t/will-libcanard-or-libuavcan-becomes-a-kind-of-software-bus-for-ipc/1119

While this is a minor change, its implications are significant. As I attempted to illustrate in https://github.com/UAVCAN/pyuavcan/pull/154, proper integration between node registers and the parameter passing services offered by the operating system (which in this case means environment variables) allows one to easily construct software nodes without having to hard-code any UAVCAN-related parameters. Currently, as can be seen in the old demo for PyUAVCAN, it is necessary to either hard-code protocol-related parameters such as node-ID or port-ID (which is non-compliant), or one has to resort to non-standard ad-hoc solutions to make them configurable. Together these factors make UAVCAN significantly less appealing outside of deeply-embedded applications than it could be. This change request provides a way to fix that.

Since the capability to map environment variables to registers can be seen critical for any non-baremetal use of UAVCAN, I would like to see it supported in other higher-level implementations beside PyUAVCAN; specifically, libuavcan (@thirtytwobits) and uavcan.rs (@davidlenfesty). The aforementioned change request to PyUAVCAN builds a case in favor of this point.

A typical embedded or baremetal application will also benefit from using a higher-level implementation than what we can offer now --- which is only Libcanard at the moment. Considering that the register API is a much more critical part of UAVCAN in v1 than it was in v0, such higher-level implementations should treat registers as the first-class feature that is automatically introduced into the user's application by the implementation library, along with its tight integration with all protocol functionality (such as automatic look-up of port-IDs from the registry at port initialization time). Again, the PyUAVCAN CR will focus on the practical aspects.

Besides that, I also want to extend the set of standard registers with two CAN-specific values: `uavcan.can.bitrate` (which is already present in some v0-based devices, this change makes it regulated), and `uavcan.can.iface` (for software nodes only).

There are some insignificant formatting changes in the doc; in particular, I added sections to improve readability.

### DIFF
--- a/uavcan/register/384.Access.1.0.uavcan
+++ b/uavcan/register/384.Access.1.0.uavcan
@@ -110,7 +110,7 @@
 #   REGISTER NAME PATTERN                               TYPE            FLAGS                   RECOMMENDED DEFAULT
 # =====================================================================================================================
 #
-#   uavcan.node.id                                      natural16       mutable, persistent     65535 (unset/PnP)
+#   uavcan.node.id                                      natural16[1]    mutable, persistent     65535 (unset/PnP)
 #
 # Contains the node-ID of the local node. Values above the maximum valid node-ID for the current transport
 # indicate that the node-ID is not set; if plug-and-play is supported, it will be used by the node to obtain an
@@ -126,7 +126,7 @@
 #
 # ---------------------------------------------------------------------------------------------------------------------
 #
-#   uavcan.pub.PORT_NAME.id                             natural16       mutable, persistent     65535 (unset, invalid)
+#   uavcan.pub.PORT_NAME.id                             natural16[1]    mutable, persistent     65535 (unset, invalid)
 #   uavcan.sub.PORT_NAME.id                             ditto           ditto                   ditto
 #   uavcan.cln.PORT_NAME.id                             ditto           ditto                   ditto
 #   uavcan.srv.PORT_NAME.id                             ditto           ditto                   ditto
@@ -161,6 +161,41 @@
 # These registers are set by the vendor once and typically they are to remain unchanged (hence "immutable").
 # The "PORT_NAME" defines the human-friendly name of the port as specified above.
 # For example, a register named "uavcan.pub.measurement.type" may contain "uavcan.si.sample.angle.Quaternion.1.0".
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.diagnostic.*
+#
+# Prefix reserved for future use.
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.can.bitrate                                  natural32[2]    implementation-defined  implementation-defined
+#   uavcan.can.iface                                    string          mutable, persistent     implementation-defined
+#
+# These registers are only relevant for nodes that support UAVCAN/CAN.
+#
+# uavcan.can.bitrate defines the CAN bus bit rate: the first value is the arbitration bit rate, the second is the
+# data phase bit rate. Nodes that support only Classic CAN should ignore the second value. Nodes that support CAN FD
+# should initialize in the Classic CAN mode (MTU 8 bytes, BRS flag not set) if the values are equal. If CAN bitrate
+# is not configurable or is always auto-detected, this register may be omitted or made immutable; otherwise it should
+# be mutable and persistent.
+#
+# uavcan.can.iface is only relevant for software nodes or nodes that are capable of using different CAN interfaces.
+# The value is a space-separated list of CAN interface names to use. The name format is implementation-defined
+# (for example, "can0").
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.udp.*
+#
+# Prefix reserved for future use.
+#
+# ---------------------------------------------------------------------------------------------------------------------#
+#
+#   uavcan.serial.*
+#
+# Prefix reserved for future use.
 #
 # ---------------------------------------------------------------------------------------------------------------------
 
--- a/uavcan/register/384.Access.1.0.uavcan
+++ b/uavcan/register/384.Access.1.0.uavcan
@@ -1,8 +1,11 @@
-# This service is used to write and read a register. Write is optional, it is performed if the value provided in
-# the request is not empty.
+# Registers are strongly-typed named values used to store the configuration parameters of a node.
+# This service is used to write and read a register.
+#
+#
+#   READ/WRITE BEHAVIORS
 #
 # The write operation is performed first, unless skipped by sending an empty value in the request.
-# The server may attempt to convert the type of the value to the proper type if there is a type mismatch
+# The server may attempt to convert the type of the supplied value to the correct type if there is a type mismatch
 # (e.g. uint8 may be converted to uint16); however, servers are not required to perform implicit type conversion,
 # and the rules of such conversion are not explicitly specified, so this behavior should not be relied upon.
 #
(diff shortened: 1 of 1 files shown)


uavcan/register/384.Access.1.0.uavcan
```diff
@@ -15,33 +18,66 @@
...
+#   - Low line (underscore) (_)
+# With the following limitations/recommendations:
+#   - The name shall not begin with a decimal digit (0-9).
+#   - The name shall neither begin nor end with a full stop.
+#   - A low line shall not be followed by a non-alphanumeric character.
+#   - The name should contain at least one full stop character.
+# Other patterns and ASCII characters are reserved for special function registers (introduced below).
+#
+#
+#   ENVIRONMENT VARIABLES
```
thirtytwobits, line 51: I really hate encouraging environment variables as a primary mechanism. It's fine as a secondary but I'd rather we define a configuration file standard as the primary mechanism. Can we not just use the same mapping but encourage a file of `KEY=VALUE\n` as the way to specify each?

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -56,21 +92,24 @@
...
+#
 # The following table specifies the register name patterns that are reserved by the specification for
 # common functions. These conventions are not mandatory to follow, but implementers are recommended to adhere because
 # they enable enhanced introspection capabilities and simplify device configuration and diagnostics.
 #
 #   REGISTER NAME PATTERN                               TYPE            FLAGS                   RECOMMENDED DEFAULT
 # =====================================================================================================================
 #
-#   uavcan.node.id                                      natural16       mutable, persistent     65535 (unset/PnP)
+#   uavcan.node.id                                      natural16[1]    mutable, persistent     65535 (unset/PnP)
```
thirtytwobits, line 112: what's this change about?

uavcan/register/384.Access.1.0.uavcan
```diff
@@ -123,18 +162,49 @@
...
+# Prefix reserved for future use.
+#
+# ---------------------------------------------------------------------------------------------------------------------
+#
+#   uavcan.can.bitrate                                  natural32[2]    implementation-defined  implementation-defined
+#   uavcan.can.iface                                    string          mutable, persistent     implementation-defined
+#
+# These registers are only relevant for nodes that support UAVCAN/CAN.
+#
+# uavcan.can.bitrate defines the CAN bus bit rate: the first value is the arbitration bit rate, the second is the
```
thirtytwobits, line 177: do we not need sample points as well?

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: For clarity. There is no way to represent a standalone scalar using the register interface. We use one-element arrays for that, and this change is intended to make it clearer.

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: IIRC, NXP were going to give this question a look in the context of DS-015 but it is yet to happen. For now, assume that the device is able to configure the timings appropriately automatically (this assumption is made by the [PythonCAN API](https://pyuavcan.readthedocs.io/en/latest/.apidoc_generated/pyuavcan.transport.can.media.pythoncan.html) in PyUAVCAN, for instance). Later we will probably need to add a dedicated register for sample point configuration.

@PetervdPerk-NXP

uavcan/register/384.Access.1.0.uavcan
pavel-kirienko: Environment variables are not meant to be directly used by the human integrator/developer excepting some unusual circumstances that require low-level access. They are convenient because they are present virtually everywhere and are familiar to virtually everybody, so it is a solid foundation to build upon. What you are describing can be implemented *on top* of the environment variable conventions that we define here. For instance, the configuration file-like behavior can be obtained using shell scripts or the orchestrator that you are already familiar with.

```sh
# my-configuration-file
UAVCAN__NODE__ID=42
UAVCAN__UDP__IFACE=127.9.0.0
UAVCAN__SUB__TEMPERATURE_SETPOINT__ID=2345
UAVCAN__SUB__TEMPERATURE_MEASUREMENT__ID=2346
UAVCAN__PUB__HEATER_VOLTAGE__ID=2347
UAVCAN__SRV__LEAST_SQUARES__ID=123
UAVCAN__DIAGNOSTIC__SEVERITY=2
```

```
$ . my-configuration-file
$ ./run_this
```

uavcan/register/384.Access.1.0.uavcan
PetervdPerk-NXP: I do concur with @thirtytwobits about having a sample point option as well. Or at least thinking about minimal sample point or something like that. MCU's can indeed calculate timings (e.g. https://github.com/apache/incubator-nuttx/blob/master/arch/arm/src/s32k1xx/s32k1xx_flexcan.c#L393). However only having the timing from just the bitrate isn't enough you might end up with a low sample point e.g. 60% which is non-ideal.

Regarding DS-015 we participated in the network topology discussion and stated if that if there's a well defined configuration of nodes and it's topology we could help analyzing the impact of different bitrates on such a topology.

pavel-kirienko: Moved this into a separate issue: https://github.com/UAVCAN/public_regulated_data_types/issues/110 any help from NXP would be appreciated.

uavcan/register/384.Access.1.0.uavcan
thirtytwobits: environment variables infer a shell environment. We are creating a dependency on the presence of a shell.

pavel-kirienko: Yes, this is by design:

> This section applies only to software nodes executed in a high-level operating system that supports environment variables or an equivalent mechanism.

A deeply embedded system is unlikely to source its configuration from a shell or a generic configuration file, so it doesn't seem relevant for that use case.

thirtytwobits: I don't know what to say. I think it's a bad design choice. You can override my opinion if you disagree.

pavel-kirienko: But if shell dependency is intentional and user-friendliness is not relevant, what makes it bad?

thirtytwobits: 1. It's an abuse of environment variables since you are basically treating it as a list. If anything it should be a delineated value within a single environment variable to be consistent with things like $PATH
2. Not all build automation tools support environment variables but they all support files.

pavel-kirienko: 1. The objective is to assign registers selectively such that each register is treated as an independent entity. A node process that is unaware of register X would not read the corresponding environment variable, so unless I misunderstood what you mean we don't seem to be following a list-like semantics here.

2. This is meant to be a method of integration-time/runtime configuration management. What build tools are you referring to and why is this relevant?

thirtytwobits (commented):
If it's runtime then this is even more inappropriate.

1. For most embedded systems there is no shell.
2. If you were developing a GUI environment variables are not appropriate. I know on Windows this has been a pain in the past where some GUI needed environment variables and you had to go through arcane and painful steps to add these to the user environment before you launched the application.

thirtytwobits (commented):
If it's runtime then this is even more inappropriate.

1. For most embedded systems there is no shell.
2. If you were developing a GUI environment variables are not appropriate. I know on Windows this has been a pain in the past where some GUI needed environment variables and you had to go through arcane and painful steps to add these to the user environment before you launched the application.

pavel-kirienko:
Deeply embedded systems and GUI software are not the applications that this feature is intended for. Instead, it is aimed at software nodes executed in a relatively high-level/POSIX-like OS (e.g., NuttX or GNU/Linux running on an embedded computer on the vehicle). The first kind (deep embedded) is likely to rely on very application-specific means of configuration and the latter can just request the user to enter the required parameters explicitly. This is to be understood as a way to instantiate node processes with specific configuration per-instance.

thirtytwobits (approved):
FIne. I don't know how many ways I can say I don't like the design but it's your call here.

Issue #112, opened 2021-04-08
reg.drone.service.battery.Parameters.0.2 is missing the nominal cell voltage


cisvanmierlo:
The specification of joule in this message is ambiguous because it is up to the implementer to use a voltage (current battery voltage, nominal voltage, max battery voltage (number of cells * max cell voltage)).

Energy and full energy could use different voltages in the joule calculation, thus this can't be divided to get the state of charge.

pavel-kirienko:
The energy specification does not say anything about voltages and it should not be expected to. It requires only that the node is to publish its best guess energy estimates:

https://github.com/UAVCAN/public_regulated_data_types/blob/53a7dbbf85554e11517c13ad91b9efb871599ad1/reg/drone/service/battery/_.0.1.uavcan#L47

Then it is up to the implementer to use the correct method of energy estimation. These are the respective definitions of **energy** and **full energy**:

> A pessimistic estimate of the amount of energy that can be reclaimed from the source in its current state.

> A pessimistic estimate of the amount of energy that can be reclaimed from a fresh source (fully fueled generator or a fully charged battery) under the current conditions (SoH, temperature, load profile, etc).

Questions of voltage or charge estimation are outside of their scope.

Then, the [state of charge](https://en.wikipedia.org/wiki/State_of_charge), understood narrowly as the amount of retrievable energy normalized into [0, 1], is found as `energy/full_energy`, without the need to obtain the electric charge. If this method is believed to produce an unreliable estimate, then the implementation is, by the definitions given above, defective.

PetervdPerk-NXP:
Hi Pavel,

Let me sketch the problem me and @cisvanmierlo is facing here.
We're implementing the DS-015 reg.drone.service.battery energy_source subject which is defined in reg.drone.service.electricity.SourceTS.

Our BMS reports full and remaining capacity in Ah, thus for SourceTS we've to convert it to Joule. To convert Ah to Joule we've to use the following formula:
<img src="https://render.githubusercontent.com/render/math?math=J^{remaining} = Ah^{remaining} * V^{pessimistic} * 3600">
Problem here is to pick which value for V? We could be extremely pessimistic and take minimum cell voltage for a LiPo e.g.
<img src="https://render.githubusercontent.com/render/math?math=V^{pessimistic} = V^{cell\_minimum} * no\_cells">
[...]

PR #113, opened 2021-04-09
reg.drone Battery parameters add nominal voltage

PetervdPerk-NXP:
As discussed in #112 
Adds the nominal voltage (the voltage that's typically on battery pack) to the reg.drone.battery.parameters message.

Battery Pack

```
uavcan.si.unit.voltage.Scalar.1.0 nominal_voltage
# The average voltage(not per-cell) the battery outputs when charged.
```

Deprecates Parameters.0.2

### DIFF
--- a/reg/drone/service/battery/Parameters.0.3.uavcan
+++ b/reg/drone/service/battery/Parameters.0.3.uavcan
@@ -6,7 +6,6 @@
 #
 # All parameters are required unless specifically stated otherwise.
 # For non-rechargeable batteries all "charge_*" parameters should be NaN.
-@deprecated
 
 truncated uint64 unique_id
 # A statistically unique number that can be used to identify this exact battery for logging and diagnostic purposes.
@@ -65,5 +64,8 @@ void1
 Technology.0.1 technology
 # The battery technology information may be leveraged by the charger to choose the appropriate charging strategy.
 
+uavcan.si.unit.voltage.Scalar.1.0 nominal_voltage
+# The average voltage(not per-cell) the battery outputs when charged.
+
 @assert _offset_.count == 1     # It is intended to be a fixed-size type (although it's not required).
-@extent 8 * 63                  # Limit to a single-frame transfer over CAN FD.
+@extent 8 * 67
--- a/reg/drone/service/battery/Parameters.0.3.uavcan
+++ b/reg/drone/service/battery/Parameters.0.3.uavcan
@@ -65,7 +65,9 @@ Technology.0.1 technology
 # The battery technology information may be leveraged by the charger to choose the appropriate charging strategy.
 
 uavcan.si.unit.voltage.Scalar.1.0 nominal_voltage
-# The average voltage(not per-cell) the battery outputs when charged.
+# The nominal voltage of the battery pack (not per-cell) as defined by the vendor.
+# The Smart Battery Data Specification v1.1 uses the term "design voltage".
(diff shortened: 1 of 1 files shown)


pavel-kirienko (approved):

echoGee:
To be picky about this , it may be a bit more clear to mention the voltage at 50% SOC ? The voltage drop is not linear from 100-0% SOC and a defining a midpoint at SOC may be more accurate, than (max-min)/2.

PetervdPerk-NXP:
But voltage at 50% State of Charge isn't the nominal voltage.

For example LiPo batteries don't have a linear discharge curve, the nominal voltage is the average voltage from this curve for example from 4.2V (Full) to 3.0V (Critical) and then you the calculate the average of the area. 

Which for LiPo is typically 3.7V.

I don't have a LiPo discharge plot at hand, but this source gives you a nice picture 

afbeelding
Source: https://www.labatteria.co.il/en/e-bike-battery-discharge-rate-4c-vs-3c/

echoGee:
> But voltage at 50% State of Charge isn't the nominal voltage.
> 
> For example LiPo batteries don't have a linear discharge curve, the nominal voltage is the average voltage from this curve for example from 4.2V (Full) to 3.0V (Critical) and then you the calculate the average of the area.
> 
> Which for LiPo is typically 3.7V.
> 
> I don't have a LiPo discharge plot at hand, but this source gives you a nice picture
> 
> afbeelding
> Source: https://www.labatteria.co.il/en/e-bike-battery-discharge-rate-4c-vs-3c/

@PetervdPerk-NXP : Is "nominal voltage is the integral under the 1C discharge curve" an accepted definition? I couldn't find that definition. Also, do you mean integral divided by the capacity ?

reg/drone/service/battery/Parameters.0.3.uavcan
```diff
@@ -0,0 +1,71 @@
...
+
+uint7 state_of_health_pct  # [percent]
+# The SoH of the battery, or best guess thereof; ranges from 0 (unusable) to 100 (new).
+void1
+
+Technology.0.1 technology
+# The battery technology information may be leveraged by the charger to choose the appropriate charging strategy.
+
+uavcan.si.unit.voltage.Scalar.1.0 nominal_voltage
+# The average voltage(not per-cell) the battery outputs when charged.
```
pavel-kirienko, line 68: Suggestion based on the conversation at #112

```suggestion
# The nominal voltage of the battery pack (not per-cell) as defined by the vendor.
# The Smart Battery Data Specification v1.1 uses the term "design voltage".
# E.g., a typical 22S LiCoO2 pack might report 81.4 V here.
```

echoGee: This, IMHO, is not a good way to define a standard parameter. lack of clarity.

pavel-kirienko: While we're waiting for the SBS Forum to respond, we could merge this tentatively to unblock @PetervdPerk-NXP and then go back and provide a proper definition later.

echoGee: Not sure I understand. Why would a nominal voltage (the definition of which is not known) block @PetervdPerk-NXP ?

pavel-kirienko: This is discussed in the beginning of #112

PR #124, opened 2021-09-17
Type signature description

flytrex-vadim:


reg/udral/README.md
```diff
@@ -236,6 +236,27 @@ It is not necessary to include the leading and trailing name component separator
 Registers that are not known to the auto-configuration authority are to be silently ignored,
 which permits addition of implementation-defined ports without breaking compatibility with standard services.
 
+
+#### Type signature register
+
+In many use cases it is desirable to have the ability to statically check the compatibility between sender and 
+receiver of a message. This capability could be used both during automatic configuration and by manual configurator tools.
+The design allows this by adding an extra register to every service pub/sub group that contains a string _message type signature_.
```
pavel-kirienko, line 244:
```suggestion
The design allows this by adding an optional register per port that contains a string _data type signature_.
```

There are message types and service types. I expect your signature is applicable to both, is it not?

reg/udral/README.md
```diff
@@ -236,6 +236,27 @@ It is not necessary to include the leading and trailing name component separator
...
+
+    uavcan.sub.main_drive.setpoint.type_sig
```
pavel-kirienko, line 246:
```suggestion
    uavcan.pub.PORT_NAME.type_sig
    uavcan.sub.PORT_NAME.type_sig
    uavcan.cln.PORT_NAME.type_sig
    uavcan.srv.PORT_NAME.type_sig

Where `PORT_NAME` is a placeholder defined in the
[documentation for `uavcan.register.Access`](https://github.com/UAVCAN/public_regulated_data_types/blob/b02e6899a319ddefbab41b820d167c95dd00174d/uavcan/register/384.Access.1.0.uavcan#L136-L139).
For example: `uavcan.sub.main_drive.setpoint.type_sig`
```

Would it not make sense to future-proof this by incorporating a version number into the register name, like `sig1`? Or am I overthinking it?

reg/udral/README.md
```diff
@@ -236,6 +236,27 @@ It is not necessary to include the leading and trailing name component separator
...
+Example signature would look like `(u12{u8}a5[f16]=i8)` , and be interpreted as
+
+ - `()` delimit sealed type
+ - `{}` delimit extensible type
+ - `b`, `u`, `i`, `f`, `v` followed by size in bits encode primitive types
+ - `a` and `A` followed by length in items and brackets encode variable and fixed size arrays
+ - `=` marks byte alignment point
+
+Sample code that generates type string from DSDL and compares two strings for compatibility can be seen at this 
+(temporary) [location](https://gitlab.com/vadimz1/dsdl-sig-gen/-/tree/main).
```
pavel-kirienko, line 257: Embedding links into text like this makes the text unintelligible when copy-pasted.

```suggestion
(temporary) location: <https://gitlab.com/vadimz1/dsdl-sig-gen/-/tree/main>.
```

reg/udral/README.md
```diff
@@ -236,6 +236,27 @@ It is not necessary to include the leading and trailing name component separator
...
+
+#### Type signature register
+
+In many use cases it is desirable to have the ability to statically check the compatibility between sender and 
+receiver of a message. This capability could be used both during automatic configuration and by manual configurator tools.
+The design allows this by adding an extra register to every service pub/sub group that contains a string _message type signature_.
+
+    uavcan.sub.main_drive.setpoint.type_sig
+
+Example signature would look like `(u12{u8}a5[f16]=i8)` , and be interpreted as
```
pavel-kirienko, line 248:
```suggestion
Example signature would look like `(u12{u8}a5[f16]=i8)`, and be interpreted as
```

A slightly more formal specification would be nice.

pavel-kirienko (changes requested):

reg/udral/README.md
flytrex-vadim: Not yet. It should be easy to add (by adding a separator between request and reply part of the string and match both with AND condition. Tagged union implementation is not complete either.

reg/udral/README.md
flytrex-vadim: Sure, I'm just reluctant to spend more time than strictly necessary right now until we see some traction.

reg/udral/README.md
flytrex-vadim: > Embedding links into text like this makes the text unintelligible when copy-pasted.

Hmm, not sure what you are saying here - the  []() format was exactly the same as you used above with github link.

reg/udral/README.md
```diff
@@ -236,6 +236,34 @@ It is not necessary to include the leading and trailing name component separator
 Registers that are not known to the auto-configuration authority are to be silently ignored,
 which permits addition of implementation-defined ports without breaking compatibility with standard services.
 
+
+#### Type signature register
+
+In many use cases it is desirable to have the ability to statically check the compatibility between sender and
```
thirtytwobits, line 242: A bit of a nit but also a question, "statically check" generally means compile-time. We're adding a runtime mechanism here. Should this be "dynamically check" or just "check" or am I not understanding the proposal?

reg/udral/README.md
```diff
@@ -236,6 +236,34 @@ It is not necessary to include the leading and trailing name component separator
...
+    uavcan.pub.PORT_NAME.type_sig
+    uavcan.sub.PORT_NAME.type_sig
+    uavcan.cln.PORT_NAME.type_sig
+    uavcan.srv.PORT_NAME.type_sig
+
+Where `PORT_NAME` is a placeholder defined in the
+[documentation for `uavcan.register.Access`](https://github.com/UAVCAN/public_regulated_data_types/blob/b02e6899a319ddefbab41b820d167c95dd00174d/uavcan/register/384.Access.1.0.uavcan#L136-L139).
+For example: `uavcan.sub.main_drive.setpoint.type_sig`
+
+Example signature would look like `(u12{u8}a5[f16]=i8)`, and be interpreted as
```
thirtytwobits, line 255: Let's note that this is a draft clause in the standard and that, once adopted, a formal grammar will be specified and breaking changes may occur.

reg/udral/README.md
```diff
@@ -236,6 +236,34 @@ It is not necessary to include the leading and trailing name component separator
...
+
+ - `()` delimit sealed type
+ - `{}` delimit extensible type
+ - `b`, `u`, `i`, `f`, `v` followed by size in bits encode primitive types
+ - `a` and `A` followed by length in items and brackets encode variable and fixed size arrays
+ - `=` marks byte alignment point
+
+Sample code that generates type string from DSDL and compares two strings for compatibility can be seen at this
```
thirtytwobits, line 263: It seems like this should, if adopted, be included as an addendum to the core DSDL standard (sort of the JSON to DSDL's JavaScript, if you will). It would also be useful for self-describing data storage of log data.

thirtytwobits (approved):

reg/udral/README.md
flytrex-vadim: It is indeed a runtime check that is intended to happen once (as opposed to on every message transmission) during system initialization (or even configuration). 
Do you have any wording suggestion that would improve on "statically" and still express this distinction?

reg/udral/README.md
flytrex-vadim: Ok

reg/udral/README.md
pavel-kirienko: >"statically check" generally means compile-time

I agree with Vadim here -- I am reading "static" as a "check that is performed before the system is operational", so it is not a runtime check, strictly speaking.

reg/udral/README.md
pavel-kirienko: My point is that the text should be meaningful when copy-pasted without the embedded link. Ex: "can be seen at this temporary location" is meaningless because the referenced location is not specified, "can be seen at Vadim's GitLab" is fine because the location is specified even if the embedded hyperlink is lost.

pavel-kirienko (approved):

PR #125, opened 2021-09-18
Replace DS-015 with UDRAL

pavel-kirienko:
Find the context and submit high-level feedback at https://forum.uavcan.org/t/first-draft-of-the-udral-dsdl-namespace/1438


auturgy:
Thanks @flytrex-vadim for your efforts in proposing a solution to type safety, with port-ID tables and type signatures.  @pavel-kirienko beyond that (which is just a note in the readme) I don't see how this PR - which is essentially find/replace drone/udral - addresses the remaining issues detailed here: https://forum.uavcan.org/t/meeting-minutes-july-21-2021-utc-udral-call/1365 and https://forum.uavcan.org/t/problems-with-ds-015/1219/47

pavel-kirienko:
@auturgy you will find a detailed review of the resolved and outstanding issues in the forum thread I linked in the OP post. Feel free to comment on them in that thread.

coderkalyan:
> This changeset is expected to be merged by Oct 1st unless there are major breaking proposals.

I would suggest allocating more time for interested parties to actually take a look at it in depth and comment - since there were concerns with the original DS-015 message set.

auturgy:
@pavel-kirienko as you well know, that post identifies and addresses issues as you choose, not as or how they've been raised by the stakeholders.
Frankly your approach to this problem raises serious concern regarding leadership and trust. In simple terms, if you don't listen to the Consortium, there won't be a Consortium.
I've raised a consortium vote in an attempt to persuade you to have these discussions publicly.

pavel-kirienko:
@auturgy if you happen to have any constructive criticism, it would be most welcome on the forum.

auturgy:
My forum feedback:

The issues raised here: https://forum.uavcan.org/t/meeting-minutes-july-21-2021-utc-udral-call/1365
and https://forum.uavcan.org/t/uavcan-drone-application-layer-sig-guidelines/1280/16 provide a good summary of the requirements a UDRAL proposal should meet. The definitions in the PR are nothing but find/replace drone/udral on the DS-015 definitions. With no attempt to address the shortfalls identified in that message set, I see no reason to expect different feedback on them.
As I’ve mentioned before, I think we need to get the design agreed and the tooling in place before publishing message definitions. I don’t think we’re at that point yet.
I’d also suggest that as definitions are developed we leave them in a branch until they’re iterated, agreed, and stable. MAVLink has demonstrated many times that WIP messages have a habit of making their way into production systems, making it very difficult to iterate and change them. That’s why in MAVLink we’ve moved away from WIP messages in favour of development.xml (which is annoying, but a necessary compromise given the workflows in that project).

.vscode/spellright.dict
```diff
@@ -406,3 +406,8 @@ CAS
 OOP
 alignee
 UML
+UDRAL
+multirotors
+DRone
```
auturgy, line 411: typo. Drone

pavel-kirienko: This is intentional as part of "UAVCAN **DR**one Application Layer (UDRAL)".

pavel-kirienko:
James, I do value your feedback, but let us please keep the conversation on the forum.

davidbuzz:
I get "Oops! That page doesn’t exist or is private." On the forum links, can't review.

PR #149, opened 2025-04-22
file/directory crc

ot-goegelem:
In order to check the integrity of file transfers or directory listings it would be nice to have a way of requesting a checksum over a file/directory in order to detect edge cases where the file/directory changes between subsequent requests.

Additionally it would also allow to do some prior checks in order to not sync any files that are already present.
 
 To be fair the cyphal specification mentions the methodology to put "hash-files" near the real files (like `image.bin.sha256`) which works for "static" files, but this becomes cumbersome for files that change all the time, because someone has to update those "hash-files". 
 Unfortunately this will also not help in the "list-directory" case as it is very hard to detect an "a file prior to the current index was deleted" error where the file that would have been at the current index prior to deletion will not show up in the listing.

### DIFF
--- a/uavcan/file/410.Crc.0.1.dsdl
+++ b/uavcan/file/410.Crc.0.1.dsdl
@@ -1,0 +1,51 @@
+# Request to compute the CRC of a file or directory.
+#
+# The CRC is computed using the CRC-32 algorithm.
+# The CRC is computed over the contents of the file or directory, not over the metadata.
+#
+# Case File:
+#   - If the path points to a file, the CRC is computed over the contents of the file.
+
+# Case Directory:
+#   - If the path points to a directory, the CRC is computed over a concatenation of the filenames.
+#   - The filenames are concatenated with a newline character (\n).
+#   - The filenames only contain the base name of the file, not the full path.
+#   - The content of sub-directories are not included in the CRC computation.
+#
+#   - Example:
+#      - Filesystem
+#        dir
+#        ├── sub-dir
+#        │   ├── sub-file1.txt
+#        │   └── sub-file2.txt
+#        ├── file1.txt
+#        └── file2.txt
+#
+#	   - The CRC request on the directory 'dir' will compute the CRC over the following content:
+#        """
+#            sub-dir
+#            file1.txt
+#            file2.txt
+#		 """
+#        i.e.: "sub-dir\nfile1.txt\nfile2.txt"
+# 	   - via python
+#       ``` python
+#           import binascii
+# 		    string = "sub-dir\nfile1.txt\nfile2.txt"
+# 		    print(f"0x{binascii.crc32(bytes(string, encoding='ascii')):08x}")
+#       ```
+#      - The CRC value will be 0x58c8a463
+
+Path.2.0 path
+
+@extent 300 * 8
+
+---
+
+Error.1.0 error
+# Result of the operation.
+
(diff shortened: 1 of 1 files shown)


uavcan/file/410.Crc.0.1.dsdl
```diff
@@ -0,0 +1,56 @@
...
+# The CRC is computed over the contents of the file or directory, not over the metadata.
+#
+# Case File:
+#   - If the path points to a file, the CRC is computed over the contents of the file.
+
+# Case Directory:
+#   - If the path points to a directory, the CRC is computed over a concatenation of the filenames.
+#   - The filenames are concatenated with a newline character (\n).
+#   - The filenames only contain the base name of the file, not the full path.
+#   - The content of sub-directories are not included in the CRC computation.
```
pavel-kirienko, line 13: How should the entries be ordered while computing the CRC?

pavel-kirienko (commented):
I am not sure if this use case, which is valid, warrants a new RPC type. The intention of this recommendation given in the Read service:

https://github.com/OpenCyphal/public_regulated_data_types/blob/f9f67906cc0ca5d7c1b429924852f6b28f313cbf/uavcan/file/408.Read.1.1.dsdl#L22-L25

Was that the hash/crc files do not have to exist as physical entries in directory, but rather they are to be generated on demand, so if you only have an `image.bin` and the remote client attempts to read `image.bin.sha256` (or any other hash), you just compute the hash and return its digest as if it was a file.

Does that make sense or am I missing something?

uavcan/file/410.Crc.0.1.dsdl
ot-goegelem: You are right. 
It is possible to do it via the virtual files, but that does not work for directory listings.
A way of solving this could be to also create virtual files for the directories (for `directory/` request `directory.sha256`), but I could see that that would fail in the case of the root directory. 
Maybe for directories this approach can be modified to request a `.<hash>` file within the directory instead (for `directory/` request `directory/.sha256`.

One objection to this and why I proposed to add another RPC for this is to make it more explicit.
With this new RPC there are no virtual files and it is clear that whenever someone calls this endpoint the server has to do some work to generate the result. In the case of the virtual files it is not clear to the user if the requested file is up-to-date or if the server returned an outdated "real file" in the same location.
Also I would argue that if someone implements a new version of the spec they will probably rather implement the  endpoint (which is clearly defined as one of the functions of a file server) instead of the on demand file generation (which is vaguely defined and is highly optional). Also with the RPC there is one specific algorithm defined (which is already used in other parts of the spec) and (if i know the devices are up-to-date with the latest spec) someone can expect to just use this endpoint. Whereas in the virtual file case it might not be clear **IF** and also **WHAT** hashes/crc the respective server implements.

Regarding the entry order in the directory case the order should be specified.
I would propose to define that the order should be the same order as the `list` order.

emrainey:
There needs to be something that can help give users a reasonable certainty that a file on the server is the same as a file on the client. SHA256 is a better algo than CRC64 or CRC32 for this. The directory listing is a challenging thing to hash given the order issues and capitalization and UTF-8 and i18n. What is the user really interested in here? That all the files were transferred or that no file has been renamed?

pavel-kirienko:
>capitalization and UTF-8 and i18n

Why are these an issue?

